PR #77, opened 2015-03-13
Set the proper accepts header for a sparql response

jcoyne:
See
http://answers.semanticweb.com/questions/31906/getty-sparql-gives-a-404-if-you-pass-accept-applicationjson

### DIFF
--- a/lib/qa/authorities/getty.rb
+++ b/lib/qa/authorities/getty.rb
@@ -50,8 +50,7 @@ module Qa::Authorities
     end
 
     def request_options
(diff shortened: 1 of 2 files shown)


lib/qa/authorities/getty.rb
```diff
@@ -50,8 +50,7 @@ def find_url id
     end
 
     def request_options
-      # Don't pass a request header. See http://answers.semanticweb.com/questions/31906/getty-sparql-gives-a-404-if-you-pass-accept-applicationjson
-      { }
+      { accept: 'sparql-results+json'}
```
cbeer: I assume there's something later in the chain that transforms this to `application/sparql-results+json`?

jcoyne: Nope, you just found an error. Thanks.

PR #81, opened 2015-04-09
Extract sub-authorities to separate classes

jcoyne:
This helps pave the way for adding local authorities that are not backed
by a yaml file.

This is not backward compatible so it should be 0.5.0


lib/qa/authorities/authority_with_sub_authority.rb
```diff
@@ -1,12 +1,18 @@
 module Qa::Authorities
-  class AuthorityWithSubAuthority < Base
```
terrellt: This is a pretty big breaking change isn't it? I don't think we have any classes that were dependent on this in OD, but I'm not sure what external implementation is like. Is it possible to deprecate?

jcoyne: I'm not sure it's worth it. This'll be a new minor version as the major is less than 1.  I think the messages that are left give you a good clue as to what happened if you run into it.  I don't think we should work too hard to carry forward a bad design.

terrellt:
I'm :+1: assuming it's okay to just break backwards compat at this point, seeing as we're not at 1.0.

https://github.com/projecthydra-labs/questioning_authority/blob/master/CONTRIBUTING.md#submitting-changes Whether I agree with the practice or not, contributing document says squash - please do so.

lib/qa/authorities/authority_with_sub_authority.rb
```diff
@@ -1,12 +1,18 @@
...
     # Registers the authority and its sub-authority if it has one
-    def initialize(sub_authority=nil)
-      @sub_authority = sub_authority
-      raise Qa::MissingSubAuthority, "No sub-authority provided" if sub_authority.nil?
-      raise Qa::InvalidSubAuthority, "Unable to initialize sub-authority #{sub_authority} for #{self.class.name}" unless sub_authorities.include?(sub_authority)
+    def new(sub_authority=nil)
+      raise "Initializing with as sub authority is removed. use #{self.class}.factory(#{sub_authority.inspect}) instead"
+    end
+
+    def factory(sub_authority)
```
dbrower: The name `factory` is not very clear, especially since this is a module method, and so it will be mixed into other classes.  Maybe `new_sub_authority`?

jcoyne: I think factory is a fairly well known concept in software. See http://en.wikipedia.org/wiki/Factory_method_pattern

jcoyne:
@terrellt I'm not squashing. Each of these commits stand on their own. Squashing is for when you have commits like "fix a problem with previous commit", "add a test", etc.

lib/qa/authorities/authority_with_sub_authority.rb
jcoyne: I will change it to "subauthority_for"

PR #82, opened 2015-06-01
New authority - a wrapper for FAST autocomplete service

hackmastera:
FYI: to get this working I had to modify terms_controller to account for a camelcased authority class. That's in a separate commit; happy to squash it if that's best

### DIFF
--- a/lib/qa/authorities/assign_fast/generic_authority.rb
+++ b/lib/qa/authorities/assign_fast/generic_authority.rb
@@ -1,0 +1,61 @@
+module Qa::Authorities
+  # A wrapper around the FAST api for use with questioning_authority
+  # API documentation:
+  # http://www.oclc.org/developer/develop/web-services/fast-api/assign-fast.en.html
+  class AssignFast::GenericAuthority < Base
+    attr_reader :subauthority
+    def initialize(subauthority)
+      @subauthority = subauthority
+    end
+
+    include WebServiceBase
+
+    # Search the FAST api
+    #
+    # @param [String] the query
+    # @return json results
+    def search q
+      url = build_query_url q
+      begin
+        raw_response = get_json(url)
+      rescue JSON::ParserError => error
+        Rails.logger.info "Could not parse response as JSON. Request url: #{url}"
+        return []
+      end
+      parse_authority_response(raw_response)
+    end
+
+    # Build a FAST API url
+    #
+    # @param [String] the query
+    # @return [String] the url
+    def build_query_url q
+      escaped_query = clean_query_string q
+      index = AssignFast.index_for_authority(subauthority)
+      return_data = "#{index}%2Cidroot%2Cauth%2Ctype"
+      num_rows = 20 # max allowed by the API
+      url = "http://fast.oclc.org/searchfast/fastsuggest?&query=#{escaped_query}&queryIndex=#{index}&queryReturn=#{return_data}&suggest=autoSubject&rows=#{num_rows}"
+    end
+
+    private
+
+    # Removes characters from the query string that are not tolerated by the API
+    #   See oclc sample code at
+    #   http://experimental.worldcat.org/fast/assignfast/js/assignFASTComplete.js
+    def clean_query_string q
+      URI.escape(q.gsub(/-|\(|\)|:/, ""))
+    end
+
+    def parse_authority_response(raw_response)
+      results = raw_response['response']['docs'].map do |doc|
+        index = AssignFast.index_for_authority(subauthority)
+        term = doc[index].first
+        if doc['type'] == 'alt'
+          term += ' USE ' + doc['auth']
+        end
+        { id: doc['idroot'], label: term, value: doc['auth'] }
+      end
+    end
+
+  end
+end
--- a/lib/qa/authorities/assign_fast_subauthority.rb
+++ b/lib/qa/authorities/assign_fast_subauthority.rb
@@ -1,0 +1,31 @@
+# Encapsulate information about assignFAST subauthorities
+module Qa::Authorities::AssignFastSubauthority
+
+  # Hash of subauthority names used in qa routes => 'index' used by API
+  Subauthorities = {
+    'all'        => 'suggestall',
+    'personal'   => 'suggest00',
+    'corporate'  => 'suggest10',
+    'event'      => 'suggest11',
+    'uniform'    => 'suggest30',
+    'topical'    => 'suggest50',
+    'geographic' => 'suggest51',
+    'form_genre' => 'suggest55',
+  }
+
+  # Get a list of subauthorities by name
+  #
+  # @return [Array<String>]
+  def subauthorities
+    Subauthorities.keys
+  end
+
+  # Get an API index name from an English name
+  #
+  # @param [String] English name
+  # @return [String] index name
+  def index_for_authority authority
+    Subauthorities[authority]
+  end
+
+end
--- a/app/controllers/qa/terms_controller.rb
+++ b/app/controllers/qa/terms_controller.rb
@@ -30,7 +30,7 @@ class Qa::TermsController < ApplicationController
 
   def init_authority
     begin
-      mod = authority_class.constantize
+      mod = authority_class.camelize.constantize
     rescue NameError => e
       logger.warn "Unable to initialize authority #{authority_class}"
       head :not_found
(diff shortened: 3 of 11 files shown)


lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    # Search the FAST api
+    #
+    # @param [String] the query
+    # @return json results
+    def search q
+      begin
+        @raw_response = get_json(build_query_url(q))
+      rescue JSON::ParserError => error
+        # TODO: log the url and the error
```
awead: Might want to make this an issue instead.

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+      parse_authority_response
+    end
+
+    # Build a FAST API url
+    #
+    # @param [String] the query
+    # @return [String] the url
+    def build_query_url q
+      # certain chars not tolerated by the API
```
awead: make this comment part of the documentation.

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+      index = AssignFast.get_index_for_authority(subauthority)
+      return_data = "#{index}%2Cidroot%2Cauth%2Ctype"
+      num_rows = 20 # max allowed by the API
+      url = "http://fast.oclc.org/searchfast/fastsuggest?&query=#{escaped_query}&queryIndex=#{index}&queryReturn=#{return_data}&suggest=autoSubject&rows=#{num_rows}"
+    end
+
+    private
+
+    def parse_authority_response
+      r = @raw_response
```
hackmastera: will fix this! just a shortcut from cutting-and-pasting

awead:
I gave it a quick look and it's :+1: to me.

lib/qa/authorities/assign_fast/generic_authority.rb
jcoyne: I prefer the TODO, then the person debugging the code can see "A-hah! I should check to see if there is an error happening where there is a TODO". It's like a signpost "HERE BE DRAGONS" :dragon:

lib/qa/authorities/assign_fast/generic_authority.rb
jcoyne: I disagree, this comment explains the following line of code. It's not about the method.

lib/qa/authorities/assign_fast/generic_authority.rb
jcoyne: In this case, I'd suggest passing the `@raw_response` as a parameter to the method. That would make this method more testable.

lib/qa/authorities/assign_fast_subauthority.rb
```diff
@@ -0,0 +1,31 @@
...
+    'topical'    => 'suggest50',
+    'geographic' => 'suggest51',
+    'form_genre' => 'suggest55',
+  }
+
+  # Get a list of subauthorities by name
+  #
+  # @return [Array<String>]
+  def get_authorities_keys
+    return Subauthorities.keys
```
jcoyne: The return is implied. You can remove `return`

lib/qa/authorities/assign_fast_subauthority.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  # Get an API index name from an English name
+  #
+  # @param [String] English name
+  # @return [String] index name
+  def get_index_for_authority authority
+    return Subauthorities[authority]
```
jcoyne: The return is implied. You can remove `return`

lib/qa/authorities/assign_fast_subauthority.rb
```diff
@@ -0,0 +1,31 @@
...
+  # @return [Array<String>]
+  def get_authorities_keys
+    return Subauthorities.keys
+  end
+
+  # Get an API index name from an English name
+  #
+  # @param [String] English name
+  # @return [String] index name
+  def get_index_for_authority authority
```
jcoyne: Idiomatic ruby method names don't begin with `get`, so how about calling this `index_for_authority`?

lib/qa/authorities/assign_fast_subauthority.rb
```diff
@@ -0,0 +1,31 @@
...
+    'uniform'    => 'suggest30',
+    'topical'    => 'suggest50',
+    'geographic' => 'suggest51',
+    'form_genre' => 'suggest55',
+  }
+
+  # Get a list of subauthorities by name
+  #
+  # @return [Array<String>]
+  def get_authorities_keys
```
jcoyne: Idiomatic ruby method names don't begin with `get`, so how about calling this `subauthorities` or `authorities_keys` ( I prefer the former)?

lib/qa/authorities/assign_fast/generic_authority.rb
hackmastera: Attempting to write an issue appears to be leading me to just do it now. However, I do think there should be more dragons.

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+      results = []
+      r['response']['docs'].each do |doc|
+        index = AssignFast.get_index_for_authority(subauthority)
+        term = doc[index].first
+        if doc['type'] == 'alt'
+          term += ' USE ' + doc['auth']
+        end
+        results << {id: doc['idroot'], label: term, value: doc['auth']}
```
jcoyne: please add a space after `{` and before `}`

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+      num_rows = 20 # max allowed by the API
+      url = "http://fast.oclc.org/searchfast/fastsuggest?&query=#{escaped_query}&queryIndex=#{index}&queryReturn=#{return_data}&suggest=autoSubject&rows=#{num_rows}"
+    end
+
+    private
+
+    def parse_authority_response
+      r = @raw_response
+      results = []
+      r['response']['docs'].each do |doc|
```
jcoyne: it looks like you could do `results = r['response']['docs'].map do ...` instead. (if you get rid of the `results << ` on line 52 also). I prefer this because it shows that it adds something to results on every iteration. I don't have to look and see if there's an alternate path through the block.

lib/qa/authorities/assign_fast/generic_authority.rb
awead: It helps explain what the overall method does and I don't believe YARD is going to pick that line up when it creates the documentation, which it why I was advocating that it be placed above where the user could read it if there were consulting the docs.

jcoyne: I think a comment explaining a regex is always welcome, especially for those who aren't comfortable with regexes. It also may point out a flaw in the regex you've written to a future programmer.

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    # Build a FAST API url
+    #
+    # @param [String] the query
+    # @return [String] the url
+    def build_query_url q
+      # certain chars not tolerated by the API
+      q.gsub! /-|\(|\)|:/, ""
+      escaped_query = URI.escape(q)
```
jcoyne: Lines 33 and 34 should probably be their own method `escaped_query(q)`. That will make it more testable.

awead: :+1: 
``` ruby
# Removes characters from the query string that are not tolerated by the API
def clean_query_string q
  URI.escape(q.gsub!(/-|\(|\)|:/, ""))
end
```

awead: You can also cheat and make it a private method... then you're not required to test it :sunglasses:

hackmastera: @awead haha that's exactly what I was just doing! I like this solution.

hackmastera:
Thanks for all these comments @jcoyne and @awead !! I believe I have accounted for everything.

PR #83, opened 2015-07-14
Added support for  Getty TGN (search and find). 

geekscruff:
Added support for  Getty TGN (search and find) by creating a new tgn.rb class.  Amended aat.rb to add the sub-authority name and getty.rb to deal with the sub-authority and choose the correct class. This is my first ever Hydra pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,10 @@
 /log/*.log
 /tmp
 Gemfile.lock
-Gemfile.lock
 coverage
 pkg
 spec/internal
+
+# Ignore software specific files
+/.idea
+/questioning_authority.iml
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,3 @@ Gemfile.lock
 coverage
 pkg
 spec/internal
-
-# Ignore software specific files
-/.idea
-/*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,4 @@ spec/internal
 
 # Ignore software specific files
 /.idea
-/questioning_authority.iml
+/*.iml
--- a/lib/qa/authorities/getty.rb
+++ b/lib/qa/authorities/getty.rb
@@ -3,14 +3,20 @@ require 'uri'
 module Qa::Authorities
   module Getty
     require 'qa/authorities/getty/aat'
+    require 'qa/authorities/getty/tgn'
     extend AuthorityWithSubAuthority
 
     def self.subauthorities
-      [ "aat" ]
+      [ "aat" , "tgn" ]
     end
 
-    def self.subauthority_class(_)
-      AAT
+    def self.subauthority_for(subauthority)
+      validate_subauthority!(subauthority)
+      if subauthority == 'aat'
+        AAT.new(subauthority)
+      elsif subauthority == 'tgn'
+        TGN.new(subauthority)
+      end
     end
   end
 end
--- a/lib/qa/authorities/getty.rb
+++ b/lib/qa/authorities/getty.rb
@@ -11,14 +11,14 @@ module Qa::Authorities
       [ "aat" , "tgn", "ulan" ]
     end
 
-    def self.subauthority_for(subauthority)
-      validate_subauthority!(subauthority)
-      if subauthority == 'aat'
-        AAT.new(subauthority)
-      elsif subauthority == 'tgn'
-        TGN.new(subauthority)
-      elsif subauthority == 'ulan'
-        Ulan.new(subauthority)
+    def self.subauthority_class(subauthority)
+      case subauthority
+        when 'aat'
+          AAT
+        when 'tgn'
+          TGN
+        when 'ulan'
+          Ulan
       end
     end
   end
--- a/lib/qa/authorities/getty.rb
+++ b/lib/qa/authorities/getty.rb
@@ -4,10 +4,11 @@ module Qa::Authorities
   module Getty
     require 'qa/authorities/getty/aat'
     require 'qa/authorities/getty/tgn'
+    require 'qa/authorities/getty/ulan'
     extend AuthorityWithSubAuthority
 
     def self.subauthorities
-      [ "aat" , "tgn" ]
+      [ "aat" , "tgn", "ulan" ]
     end
 
     def self.subauthority_for(subauthority)
@@ -16,6 +17,8 @@ module Qa::Authorities
         AAT.new(subauthority)
       elsif subauthority == 'tgn'
         TGN.new(subauthority)
+      elsif subauthority == 'ulan'
+        Ulan.new(subauthority)
       end
     end
   end
--- a/lib/qa/authorities/getty/aat.rb
+++ b/lib/qa/authorities/getty/aat.rb
@@ -2,11 +2,6 @@ module Qa::Authorities
   class Getty::AAT < Base
     include WebServiceBase
 
-    attr_reader :subauthority
-    def initialize(subauthority)
-      @subauthority = subauthority
-    end
-
     def search q
       parse_authority_response(json(build_query_url(q)))
     end
--- a/lib/qa/authorities/getty/aat.rb
+++ b/lib/qa/authorities/getty/aat.rb
@@ -29,7 +29,7 @@ module Qa::Authorities
                  skos:inScheme <http://vocab.getty.edu/aat/> ;
                  gvp:prefLabelGVP [skosxl:literalForm ?name].
               FILTER regex(?name, \"#{search}\", \"i\") .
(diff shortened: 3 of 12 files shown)


lib/qa/authorities/getty/tgn.rb
```diff
@@ -0,0 +1,65 @@
...
+      { accept: 'application/sparql-results+json'}
+    end
+
+    private
+
+    # Reformats the data received from the service
+    # Adds the parentString to the location name to enable disambiguation
+    def parse_authority_response(response)
+      response['results']['bindings'].map do |result|
+        { 'id' => result['s']['value'], 'label' => result['name']['value']  + ' (' + result['par']['value'] + ')' }
```
scande3: Would it make sense to replace the last two values of the parent term? "World" doesn't seem useful and the continent value likely isn't needed. This might make it easier for a user to quickly browse what results they are getting. Something like result['par']['value'].gsub(/\,[^\,]+\,[^\,]+$/, '')?

lib/qa/authorities/getty/tgn.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    def build_query_url q
+      query = URI.escape(sparql(untaint(q)))
+      "http://vocab.getty.edu/sparql.json?query=#{query}&_implicit=false&implicit=true&_equivalent=false&_form=%2Fsparql"
+
+    end
+
+    def sparql(q)
+      search = untaint(q)
```
scande3: Should more parsing be done here? I know it doesn't exist anywhere else but this current form is quite limited. For example, "Atlanta" doesn't give me the likely best match of "Atlanta, GA".

So maybe it might make sense to split the search term on either the first "," or "(" and trim that part of the string to match the gvp:prefLabelGVP. Then use the remainder to match the parent string via regex? This way one might be able to do "Atlanta, Georgia" or "Atlanta (Georgia" and get better results?

(May be a bad idea and wouldn't work for "Atlanta, GA")

scande3:
I added two possible suggestions for improvements. Let me know what you think about them. Otherwise this looks good to me for a TGN QA resolver.

geekscruff:
Brilliant, thanks. I have a new version ready to address those ... I also
noticed ULAN is now available from Getty, so will spend a bit of time
adding that too.

Julie

On 15 July 2015 at 01:07, Steven Anderson <notifications@github.com> wrote:

> I added two possible suggestions for improvements. Let me know what you
> think about them. Otherwise this looks good to me for a TGN QA resolver.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/projecthydra-labs/questioning_authority/pull/83#issuecomment-121432531>
> .
>



-- 
Julie Allinson <julie.allinson@york.ac.uk>
Manager, Digital York (supporting Library & Archives IT)
Information
University of York
Heslington, York, YO10 5DD, UK
tel: ++44 (0) 1904 324083

skype: j.allinson | gtalk: julie.allinson@york.ac.uk
orcid: http://orcid.org/0000-0001-6459-6227
calendar: http://tinyurl.com/jal-gcal

email disclaimer: http://www.york.ac.uk/docs/disclaimer/email.htm
please sponsor me to run the York marathon 2015:
https://www.justgiving.com/MatildaBScruff

geekscruff:
Hello. I added the regex suggested to strip off the Continent and World part. I also spent a bit of time improving the TGN SPARQL query. It now checks both the name and the parent string and works on the following logic:

- one term MUST appear in the name
- all terms MUST appear in either name or parent string (but NOT in the Continent and 'World)

This would mean 'Atlanta,Georgia" would now work (although Atlanta GA still won't). It leads to more results being returned, but I think it's altogether better.

It also now doesn't limit to 10 results (and I've removed that restriction from AAT also) and sorts results by name, then parent.

I've also added ULAN. If there is more than one term it searched both preferred and alt labels.

There is a bit of hacky gsub-ing in the TGN code as the url encoding within the class was failing in the getty service.

Let me know if you have any more suggestions - this is been a good learning experience!

.gitignore
```diff
@@ -15,7 +15,10 @@
 /log/*.log
 /tmp
 Gemfile.lock
-Gemfile.lock
 coverage
 pkg
 spec/internal
+
+# Ignore software specific files
```
jcoyne: These changes should not get committed. You should add these to a global gitignore. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

lib/qa/authorities/getty.rb
```diff
@@ -3,14 +3,23 @@
...
     def self.subauthorities
-      [ "aat" ]
+      [ "aat" , "tgn", "ulan" ]
     end
 
-    def self.subauthority_class(_)
-      AAT
+    def self.subauthority_for(subauthority)
+      validate_subauthority!(subauthority)
+      if subauthority == 'aat'
```
jcoyne: How about 
```ruby
case subauthority
when 'aat'
  ...
when 'tgn'
  ...
when 'ulan'
  ...
end
```

lib/qa/authorities/getty/aat.rb
```diff
@@ -1,6 +1,12 @@
 module Qa::Authorities
   class Getty::AAT < Base
     include WebServiceBase
+
+    attr_reader :subauthority
+    def initialize(subauthority)
+      @subauthority = subauthority
```
jcoyne: Why do we need this parameter, instance variable?

lib/qa/authorities/getty/aat.rb
```diff
@@ -23,7 +29,7 @@ def sparql(q)
                  skos:inScheme <http://vocab.getty.edu/aat/> ;
                  gvp:prefLabelGVP [skosxl:literalForm ?name].
               FILTER regex(?name, \"#{search}\", \"i\") .
-            } LIMIT 10"
```
jcoyne: The limit was in place to keep the query from taking too long to execute. Why are you removing it?

lib/qa/authorities/getty.rb
```diff
@@ -3,14 +3,23 @@
...
+    require 'qa/authorities/getty/tgn'
+    require 'qa/authorities/getty/ulan'
     extend AuthorityWithSubAuthority
 
     def self.subauthorities
-      [ "aat" ]
+      [ "aat" , "tgn", "ulan" ]
     end
 
-    def self.subauthority_class(_)
```
jcoyne: Please don't remove this. Use this to return the class rather than overriding subauthority_for to return an instance.

lib/qa/authorities/getty/tgn.rb
```diff
@@ -0,0 +1,85 @@
+module Qa::Authorities
+  class Getty::TGN < Base
+    include WebServiceBase
+
+    attr_reader :subauthority
+    def initialize(subauthority)
```
jcoyne: we don't need this initializer

lib/qa/authorities/getty/ulan.rb
```diff
@@ -0,0 +1,76 @@
+module Qa::Authorities
+  class Getty::Ulan < Base
+    include WebServiceBase
+
+    attr_reader :subauthority
+    def initialize(subauthority)
```
jcoyne: We don't need this initializer

lib/qa/authorities/getty/aat.rb
geekscruff: To allow for more results. Still seems very quick, AAT search for art c. 600ms. Happy to put the limit back. TGN returns more results but still reasonably performant on queries that aren't too broad.

geekscruff:
Thanks @jcoyne - I think I've addressed all those. Let me know what you think about removing the LIMIT on the SPARQL query.

lib/qa/authorities/getty/aat.rb
jcoyne: Okay, we'll leave it out then. In the future it might be something we want to add back, but via a settable attribute.

jcoyne:
@geekscruff thanks. This looks good. @mjgiarlo does @geekscruff have a CLA on file?

mjgiarlo:
@geekscruff Hi! Could you share your name and your employer's name with me? If you'd prefer to do that privately, hit me at michael@psu.edu.  We require contributor license agreements on file for all code contributions to Hydra, and I'm not familiar with your GitHub username to know whether you have those on file or not.  Thanks.

geekscruff:
No probs, Julie Allinson, University of York, UK, julie.allinson@york.ac.uk.

mjgiarlo:
@jcoyne Yes, @geekscruff is fully CLA'ed.  (Thanks, Julie!)

PR #87, opened 2015-12-02
Fix Mesh#search to filter results

phebal:
`Qa::Authorities::Mesh#search` uses a non-existing class variable `@q`
in the `WHERE` clause value. This results in the value of `%` after
interpolation and all records in the table being returned.

### DIFF
--- a/lib/qa/authorities/mesh.rb
+++ b/lib/qa/authorities/mesh.rb
@@ -3,7 +3,7 @@ module Qa::Authorities
 
     def search q
       begin
-        r = Qa::SubjectMeshTerm.where('term_lower LIKE ?', "#{@q}%").limit(10)
+        r = Qa::SubjectMeshTerm.where('term_lower LIKE ?', "#{q}%").limit(10)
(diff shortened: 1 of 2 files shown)


awead:
@phebal I think it may need an engine_cart update. If you look at the output it tells you what to do:
https://travis-ci.org/projecthydra-labs/questioning_authority/jobs/94493662#L181

phebal:
This is the best I could do for now. The tests with specific version of rails don't pass because `engine_cart` generated version is always set to `2.4.2` with the bundled version is at `2.1.8`. If there's a way to tell `engine_cart` to generate specific version of rails, I can't seem to find it and don't have the time to look more into this right now. Feel free to close and I'll re-open if I find some time to fix the engine_cart problem.

jcoyne:
@phebal engine cart has been updated on master.  Would you care to rebase this branch?

phebal:
Thanks for the heads up @jcoyne I'll give it a try today or tomorrow.

phebal:
OK, it's rebased, sorry for the delay.

PR #94, opened 2016-06-29
Store the config in a module rather than a global constant

jcoyne:


cbeer:
Is there anything we need to do to preserve backwards compatibility? I'm not sure what the downstream expectations are..

jcoyne:
I don't think so, this is all internal.  Plus VERSION < 1.0

barmintor:
if the use of that config is local to Qa::Authorities::LocalSubauthority, is there a reason not to attach it there?

cbeer:
Rebase (now that `Qa::Authorities::LocalSubauthority` is removed..)?

barmintor:
Rebased.

PR #104, opened 2016-07-13
Changing the generator to link to the name spaced table.

### LINKED ISSUE #103: rails generate qa:local:tables Does not generate the correct foreign key
The current generator creates a foreign key agains local_authority when it should be creating a foreign key against qa_local_authorities.
code
```
    authority = Qa::LocalAuthority.find_or_create_by(name: 'test')
    Qa::LocalAuthorityEntry.create(local_authority: authority,
                                         label: 'abc',
                                         uri: '123')

```

error
```
*** ActiveRecord::InvalidForeignKey Exception: Mysql2::Error: Cannot add or update a child row: a foreign key constraint fails (`scholarsphere_test`.`qa_local_authority_entries`, CONSTRAINT `fk_rails_cee742275b` FOREIGN KEY (`local_authority_id`) REFERENCES `local_authorities` (`id`)): INSERT INTO `qa_local_authority_entries` (`local_authority_id`, `label`, `uri`, `created_at`, `updated_at`) VALUES (1, 'abc', '123', '2016-07-13 14:53:10', '2016-07-13 14:53:10')
```

cam156:
Fixes #103 

I tried creating the correct code based on only the generator, but could not succeed.  So I added a gsub to add the class name.

### DIFF
--- a/lib/generators/qa/local/tables/tables_generator.rb
+++ b/lib/generators/qa/local/tables/tables_generator.rb
@@ -7,6 +7,13 @@ module Qa::Local
     def migrations
       generate "model qa/local_authority name:string:uniq"
       generate "model qa/local_authority_entry local_authority:references label:string uri:string:uniq"
+      migration_file = Dir.entries(File.join(destination_root,'db/migrate/'))
+                           .reject{|name| !name.include?('create_qa_local_authority_entries')}.first
+      migration_file = File.join('db/migrate',migration_file)
+      gsub_file migration_file,
+                't.references :local_authority, index: true, foreign_key: true',
+                't.references :local_authority, foreign_key: { to_table: :qa_local_authorities }, index: true'
+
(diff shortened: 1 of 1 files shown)


lib/generators/qa/local/tables/tables_generator.rb
```diff
@@ -6,7 +6,11 @@ class TablesGenerator < Rails::Generators::Base
 
     def migrations
       generate "model qa/local_authority name:string:uniq"
-      generate "model qa/local_authority_entry local_authority:references label:string uri:string:uniq"
+      generate "model qa/local_authority_entry qa_local_authority:references label:string uri:string:uniq"
```
jcoyne: How about just gsub in the generated migration? Then we wouldn't need to change the model.

cam156: Sure! The reason I picked the model was because I knew where it was.  Any easy way to know what files the generator created

cam156:
@jcoyne better?

PR #107, opened 2016-07-18
Adding rdf authority parser for filling local authority tables

cam156:
This PR resurrects code from ScholarSphere that parses the downloadable files from Library of Congress and Lexvo to fill in the local authority tables.

### DIFF
--- a/lib/qa/services/rdf_authority_parser.rb
+++ b/lib/qa/services/rdf_authority_parser.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+module Qa
+  module Services
+    # Class to parse downloaded content from rdf sources such as
+    # Library of Congress and Lexvo to create LocalAuthorityEntries
+    # for a named LocalAuthority.
+    #
+    # Example calls:
+    # RDFAuthorityParser.import_rdf('subjects',['/tmp/authoritiessubjects.nt'])
+    # RDFAuthorityParser.import_rdf('languages',['/tmp/lexvo_2013-02-09.rdf'],
+    #        format: 'rdfxml',
+    #        predicate: RDF::URI("http://www.w3.org/2008/05/skos#prefLabel"))
+    #
+    class RDFAuthorityParser
+      class << self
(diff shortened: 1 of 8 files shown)


lib/qa/services/rdf_authority_parser.rb
```diff
@@ -0,0 +1,60 @@
...
+        # @param [Array]  sources An array of file names to process
+        # @param [Hash]   opts    options for processing
+        # @option [opts]  :format     format of the rdf to parse
+        #                              see RDF::Reader for valid options
+        #                             defaults to :ntripples
+        # @option [opts]  :predicate  label predicate to use for
+        #                              LocalAuthorityEntry.label, which
+        #                              must match a field in the source file(s)
+        #                             defaults to http://www.w3.org/2004/02/skos/core#prefLabel
+        def harvest_rdf(name, sources, opts = {})
```
jcoyne: Can this be 'import' instead of 'harvest'?

cam156:
@jcoyne and others This seems to be finished.  Any more changes needed

qa.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
   s.add_dependency "nokogiri", "~> 1.6.0"
   s.add_dependency "activerecord-import"
   s.add_dependency "deprecation"
+  s.add_dependency "linkeddata"
```
jcoyne: This seems like a big dependency to pull in, and it may not ever be used. Can we just require the application provide whatever format readers they need?

cam156: @jcoyne What do you suggest adding it as a development dependency and then making the code check to see if it is defined before requiring it? Or removing the tests and just leave the code untested? Or close the PR and put the code in ScholarSphere?

jcoyne: I think it would be a good development dependency.  Then also move the `require 'linkeddata'` to the `spec_helper.rb`

cam156: 👌

cam156:
@jcoyne Better?

PR #110, opened 2016-08-08
Bump dependency on rails to 5 or 4.2

jcoyne:


no-reply:
Does this gem fail on Rails < 4.2 now? Or are we just yanking support preemptively across the board?

jcoyne:
@no-reply yeah, I don't think we can support versions that are not even supported by rails-core.

no-reply:
Fair.

PR #112, opened 2016-09-27
TableBasedAuthority should lazily check for indexes

cbeer:
By eagerly checking, we unnecessarily make database connections even for e.g. compiling assets.


jcoyne:
I get where you're going with this, but if it's lazy, I wonder if anyone will ever see it.

PR #116, opened 2016-10-25
Enable Rubocop

awead:


.rubocop_todo.yml
```diff
@@ -0,0 +1,23 @@
+require: rubocop-rspec
+
+Metrics/PerceivedComplexity:
+  Exclude:
+
+Metrics/ClassLength:
+  Exclude:
+
+Metrics/ModuleLength:
+  Exclude:
```
jcoyne: Any idea why are these empty?

awead: Copied from Sufia, but also I think they indicate that they go in "todo" instead of the main rubocop yml file.

PR #119, opened 2016-11-11
Add search method to see if wants the controller with the request.

andjsmit:
Needed for a local QA that uses a SearchBuilder.

### DIFF
--- a/app/controllers/qa/terms_controller.rb
+++ b/app/controllers/qa/terms_controller.rb
@@ -14,7 +14,7 @@ class Qa::TermsController < ApplicationController
 
   # Return a list of terms based on a query
   def search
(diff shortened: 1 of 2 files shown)


jcoyne:
Can you please add a test?

andjsmit:
@jcoyne I'm stuck on how to test this as no included authorities would require two arguments for its search method. Not sure how to go about mocking an authority if that is what is needed. Could you give me a clue on how to get started. Thanks.

jcoyne:
You can define an authority class  that takes two args in the test.

jcoyne:
This looks good @andjsmit. Can you squash these commits?

andjsmit:
@jcoyne Squashed it. Thanks.

PR #123, opened 2017-01-16
Make `Authorities::Base` explicitly abstract

no-reply:
Fixes a pair of TODO items in `Authorities::Base`. A minor adjustment is made to the terms controller to catch the new errors. 

Full implementations of the Authority interface should be unchanged; for partial implementations this may reveal bugs.

----

The first commit (9a7318e) adds some documentation for the `Authorities::Base` class. A few major ToDo items remain. Quoting from the commit message:

> A few details are missing, as I'm not sure the interface is actually very clearly defined. Questions:
>
>   - Is there a return type for `#find(id)`? Possibly `Hash` or
>     `#to_json`?
>   - Does `#all` return an `Enumerable<Hash/#to_json>` or `Array`?
>   - Is `#search` a required method?
>
> If I find obvious answers to these, or come up with a proposal, I'll update as needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ Gemfile.lock
 coverage
 pkg
 .internal_test_app
+
+/doc
+.yardoc
--- a/app/controllers/qa/terms_controller.rb
+++ b/app/controllers/qa/terms_controller.rb
@@ -9,7 +9,11 @@ class Qa::TermsController < ApplicationController
 
   # If the subauthority supports it, return a list of all terms in the authority
   def index
-    render json: @authority.all
+    render json: begin
+      @authority.all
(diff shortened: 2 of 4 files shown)


awead:
@no-reply thanks for these. It looks like the return type should be JSON, at least, the README states that "Results are returned in JSON" but we might want to re-think this. At the time that was written, we were assuming a web environment. We could change this and clarify that QA doesn't make any assumptions about your environment and simple return hashes or arrays, etc. and the consuming application can call `.to_json` on it if they need to.

Regarding `#search`, it is not required because not all authorities support it. But, it could be required in the base class and return a NotImplemented error by default indicating that it's only supported if the authority supports it.

PR #125, opened 2017-01-17
Deprecate `WebServiceBase#get_json`; use `#json`

no-reply:
Many implementations already overrode `#get_json` with `#json`, since the former is not idiomatic. This simply deprecates the `#get_json` option in favor of a base `#json`.


no-reply:
Force pushed, minus the whitespace.

PR #102, opened 2016-07-06
Change default Faraday parameter encoder

NedHenry:
This changes sets Faraday to use a parameter encoder that does not add any square brackets to repeated query parameters.

This is intended to solve #101 

[Here is a link to the Faraday discussion about this issue.](https://github.com/lostisland/faraday/issues/182)

Issue #101, opened 2016-07-06
Faraday adds brackets to LOC queries


NedHenry:
Modern versions of Faraday (which is used to query the LOC endpoints) automatically add brackets to repeated query parameters. In other words, 
`http://id.loc.gov/search/?q=query&q=cs:http://etc.gov`
becomes
`http://id.loc.gov/search/?q[]=query&q[]=cs:http://etc.gov`
This causes my queries to be misinterpreted, so I get the same response from every LOC query no matter which subauthority or query string I use. 
[Here is the faraday issue addressing this issue.](https://github.com/lostisland/faraday/issues/182)

mjgiarlo:
Closed by #129, I believe.

PR #129, opened 2017-02-02
refactors nedhenry's patch to use flat params with LOC

revgum:
Ran into the issue with needing `Faraday::FlatParamsEncoder`. See #101 
Ned's original PR #102 needed a slight change for the tests to work, and then a manual test on my code confirmed that LOC subject searching works as expected. 

QA + Faraday produces a proper looking url like: 
http://id.loc.gov/search/?format=json&q=Giant+panda--Breeding&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&

Thanks Ned!

### DIFF
--- a/spec/lib/authorities/loc_spec.rb
+++ b/spec/lib/authorities/loc_spec.rb
@@ -28,16 +28,37 @@ describe Qa::Authorities::Loc do
     end
 
     context "for searching" do
+      let(:url) { 'http://id.loc.gov/search/?q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json' }
       it "returns a url" do
-        url = 'http://id.loc.gov/search/?q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json'
         expect(authority.build_query_url("foo")).to eq(url)
       end
     end
 
     context "for returning single terms" do
+      let(:url) { "http://id.loc.gov/authorities/subjects/sh2002003586.json" }
       it "returns a url with an authority and id" do
-        url = "http://id.loc.gov/authorities/subjects/sh2002003586.json"
         expect(authority.find_url("sh2002003586")).to eq(url)
+      end
+    end
+  end
+
+  describe "#response" do
+    subject { authority.response(url) }
+    let :authority do
+      described_class.subauthority_for("subjects")
+    end
+
+    before do
+      stub_request(:get, "http://id.loc.gov/search/?format=json&q=cs:http://id.loc.gov/authorities/subjects")
+        .with(headers: { 'Accept' => 'application/json' })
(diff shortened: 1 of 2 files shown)


revgum:
Of course, sorry I overlooked that @jcoyne .. so the test ensures that the url is produced to have the flat params with two `q` on the querystring and breaks if/when the `response` method is changed on `generic_authority.rb`.

spec/lib/authorities/loc_spec.rb
```diff
@@ -26,17 +26,27 @@
...
     context "for searching" do
       it "returns a url" do
-        url = 'http://id.loc.gov/search/?q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json'
         expect(authority.build_query_url("foo")).to eq(url)
       end
     end
 
+    context "with flat params encoded" do
+      it "returns a response" do
+        stub_request(:get, "http://id.loc.gov/search/?format=json&q=cs:http://id.loc.gov/authorities/subjects")
```
jcoyne: Can you move this to a before block please?

spec/lib/authorities/loc_spec.rb
```diff
@@ -26,17 +26,27 @@
...
+    let(:url) { 'http://id.loc.gov/search/?q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json' }
 
     context "for searching" do
       it "returns a url" do
-        url = 'http://id.loc.gov/search/?q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json'
         expect(authority.build_query_url("foo")).to eq(url)
       end
     end
 
+    context "with flat params encoded" do
```
jcoyne: Can you move this to it's own describe block that is describing the `#response` method?

spec/lib/authorities/loc_spec.rb
```diff
@@ -26,17 +26,27 @@
...
+      it "returns a response" do
+        stub_request(:get, "http://id.loc.gov/search/?format=json&q=cs:http://id.loc.gov/authorities/subjects")
+          .with(headers: { 'Accept' => 'application/json' })
+          .to_return(status: 200, body: "")
+        flat_params_url = "http://id.loc.gov/search/?format=json&q=foo&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects"
+        expect(authority.response(url).env.url.to_s).to eq(flat_params_url)
```
jcoyne: move `authority.response(url)` to a subject block.  E.g.:
```ruby
subject { authority.response(url) }
```
then test with:
```ruby
expect(subject.env.url.to_s).to eq ...
```

revgum:
now 📲  @jcoyne for review

PR #131, opened 2017-02-04
Add code to process linked data authorities via configuration

elrayle:
Assumptions…
* Authority defines an HTTP API that accepts a query string and returns one or more results in a supported linked data format.
* Authority defines an HTTP API that accepts an ID or URI and returns linked data for the identified term in a supported format.
* Supported formats (at this time):  application/rdf+xml, application/json (json-ld)

Includes configurations for...
* OCLC Fast Linked Data (supports search query and term retrieval)
* FAO Agrovoc Linked Data (supports search query and term retrieval)
* Library of Congress Linked Data (supports term retrieval only)

See README for details on configuration, results format for queries and terms, and other usage information.


elrayle:
See PR #109 for original comments for this work.

NOTE: The only differences between the original work and this PR is addition of default sorting.  This applies to...
* search - results sorted by label predicate values
* term and search - sort multi-values for labels and altlabels 
  * when multiple values exist, they are sorted by the values within a label/altlabel
  * when multi-language values exist, they are sorted by language first and then by value

elrayle:
The predicate used for sorting can be configured per authority, such that, if an authority provides a predicate for search ranking, the sort can be done on that predicate.  Alpha sorting on pref label has limited usefulness as it introduces a bias toward labels starting with letters early in the alphabet.  This may not be significant if the number of search matches is small, but will likely be less than desirable as the number of matches grows.

lib/qa/authorities/linked_data/generic_authority.rb
```diff
@@ -0,0 +1,458 @@
...
+          consolidated_results = consolidate_term_results(results)
+          json_results = convert_term_to_json(consolidated_results)
+          termhash = select_json_result_for_id(json_results, id)
+
+          predicates_hash = predicates_with_subject_uri(graph, termhash[:uri])
+          termhash['predicates'] = predicates_hash unless predicates_hash.length <= 0
+          termhash
+        end
+
+        def filter_language(graph, language)
```
no-reply: Can this be refactored to avoid loops through large graphs?

The `language.nil?` check could probably just be moved outside the `#each` call. This would avoid touching every statement when we already know we won't do anything. The rest seems like it might be possible to implement as a query, which would avoid the need to delete statements.

elrayle:
@no-reply Thanks for the feedback.  I'll take a look at making adjustments.  Likely not until next week.  I'm on a sprint for another project.

elrayle:
@no-reply The changes you requested have been completed.  Can you review this PR again?

.gitignore
```diff
@@ -18,6 +18,14 @@ Gemfile.lock
 coverage
 pkg
 .internal_test_app
+.byebug_history
+
+# IDE files
+.idea
```
jcoyne: All of these changes belongs in your local global gitignore, not in the project .gitignore

https://robots.thoughtbot.com/global-gitignore

.rubocop_todo.yml
```diff
@@ -2,17 +2,25 @@ require: rubocop-rspec
 
 Metrics/PerceivedComplexity:
   Exclude:
+    - 'lib/qa/authorities/linked_data/config/config_merge.rb'
```
jcoyne: Why are we ignoring these warnings instead of refactoring away the cause of the warning?  Moreover, if ignoring is absolutely required, I'd prefer an inline ignore, so I can see exactly what method is causing this problem.

app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  # Provide a warning if there is a request for all terms.
+  def index
+    logger.warn 'Linked data authorities do not support retrieving all terms.'
+    head :not_found
+  end
+
+  # Return a list of terms based on a query
+  def search
+    subauth = subauthority
```
jcoyne: I'm not sure assigning to a local has any benefit here. There's no work being done in that method.

app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  private
+
+    def check_vocab_param
+      head :not_found unless params[:vocab].present?
+    end
+
+    def init_authority
+      begin
```
jcoyne: Extract this begin, rescue, end to its own method please.

spec/lib/authorities/linked_data/config_merge_spec.rb
```diff
@@ -0,0 +1,466 @@
+require 'spec_helper'
+require 'qa/authorities/linked_data/config/config_merge'.freeze
```
jcoyne: Why are we requiring this here? This should be loaded via the engine.

spec/lib/authorities/linked_data/config_merge_spec.rb
```diff
@@ -0,0 +1,466 @@
...
+
+describe Qa::Authorities::LinkedData::ConfigMerge do
+  before do
+    @min_config = Qa::Authorities::LinkedData::Config.new(:LOD_MIN_CONFIG)
```
jcoyne: use `let(:min_config)` rather than instance variables in tests.

spec/lib/authorities/linked_data/config_merge_spec.rb
```diff
@@ -0,0 +1,466 @@
...
+    @full_config = Qa::Authorities::LinkedData::Config.new(:LOD_FULL_CONFIG)
+  end
+
+  after do
+    LINKED_DATA_AUTHORITIES_CONFIG[:LOD_MIN_CONFIG] =
```
jcoyne: Can you add a comment explaining what this is trying to accomplish?

lib/qa/authorities/linked_data/config/config_merge.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      def merge_variable(full_var, override_var)
+        full_var[:@type] = override_var[:@type] if override_var.key? :@type
+        full_var[:property] = override_var[:property] if override_var.key? :property
+        full_var[:required] = override_var[:required] if override_var.key? :required
+        full_var[:default] = override_var[:default] if override_var.key? :default
+      end
+
+      def merge_reppatterns(full_patterns, override_patterns)
```
jcoyne: Can you give this method a more descriptive name?

lib/qa/authorities/linked_data/config/config_merge.rb
```diff
@@ -0,0 +1,76 @@
+module Qa::Authorities
+  module LinkedData
+    module ConfigMerge
+      def merge(full_cfg, override_cfg)
+        merge_term(full_cfg[:term], override_cfg[:term]) if override_cfg.key? :term
+        merge_search(full_cfg[:search], override_cfg[:search]) if override_cfg.key? :search
+      end
+
+      def merge_term(full_term, override_term)
```
jcoyne: Should all these methods be private? If not, please add YARD doc.

lib/qa/authorities/linked_data/config/config_merge.rb
```diff
@@ -0,0 +1,76 @@
+module Qa::Authorities
+  module LinkedData
+    module ConfigMerge
```
jcoyne: Add module documentation. What does this do? When do you use it? How? And a brief example would be 👍

lib/qa/authorities/linked_data/config/config_merge.rb
```diff
@@ -0,0 +1,76 @@
...
+      def merge(full_cfg, override_cfg)
```
jcoyne: Please add yard doc to this method describing the parameters and return value.

lib/qa/authorities/linked_data/config.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      # Initialize to hold the configuration for the specifed authority.  Configurations are defined in config/authorities/linked_data.  See README for more information.
+      # @param [String] the name of the configuration file for the authority
+      # @return [Qa::Authorities::LinkedData::Config] instance of this class
+      def initialize(auth_name)
+        @authority_name = auth_name
+        auth_config
+      end
+
+      class << self
```
jcoyne: Instead  of doing this, can you just `extend Qa::Authorities::LinkedData::ConfigMerge`  Please provide some documentation about what is being included and for what purpose?  I don see `merge` being called anywhere in this file. Can you help me understand why this is needed here?

lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -0,0 +1,182 @@
+module Qa::Authorities
+  module LinkedData
+    module TermConfig
```
jcoyne: Provide module documentation about what this is for and where we should be using it.  Are all of these public methods?  Could this have been a class rather than a mixin?

lib/qa/authorities/linked_data/config/search_config.rb
```diff
@@ -0,0 +1,157 @@
+module Qa::Authorities
+  module LinkedData
+    module SearchConfig
```
jcoyne: Provide module documentation about what this is for and where we should be using it.  Are all of these public methods?  Could this have been a class rather than a mixin?

.rubocop_todo.yml
elrayle: That's pretty cool.  I didn't know you could do an inline ignore for rubocop.

.gitignore
elrayle: Thanks for the documentation on how to do that.  I added a global gitignore.

lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -0,0 +1,122 @@
...
+        end
+
+        def wrap_labels(labels)
+          lbl = "" if labels.nil? || labels.size.zero?
+          lbl = labels.join(', ') if labels.size.positive?
+          lbl = '[' + lbl + ']' if labels.size > 1
+          lbl
+        end
+
+        def sort_search_results(json_results)
```
mbarnett: As a casual reader of the codebase, can I suggest adding a comment explaining the motivation behind, and logic of, what's going on in this method? It's quite complex as written, and I'm having trouble figuring out exactly what's going on, and therefore the motivation behind certain behaviours (to pick just one example, why hash `a` is automatically less-than hash `b` if `a` doesn't have a sort key, even if `b` doesn't as well).

lib/qa/authorities/linked_data/config/config_merge.rb
elrayle: It will take a bit for me to work through all your comments.  To keep the conversation going, here is the yardoc for the module...

```
# The methods of this modules merge a partial linked data configuration into a QA provided full configuration.  QA full configurations
# are defined in QA/config/authorities/linked_data.  Partial configurations are defined in apps at APP/config/authorities/linked_data.
# Configurations are read in during application initialization by QA/config/initializers/linked_data_autorities.rb.  If a partial is
# defined by an app, the initializer will call #merge.
```

lib/qa/authorities/linked_data/config/config_merge.rb
elrayle: Yardoc for the merge method...
```
# Merge a full configuration defined in QA with a partial configuration defined in an app.
# @param [Hash] Full configuration read in from QA/config/authorities/linked_data
# @param [Hash] Partial configuration overriding parts of a QA provided config, read in from APP/config/authorities/linked_data
# @example partial config to change term language to French
#   { "term": { "language": ["fr"] } }
```

lib/qa/authorities/linked_data/search_query.rb
elrayle: I agree that the sorting code was complex.  I moved the processing of special cases into separate methods and added comments.  Take a look and let me know if this addresses your concerns.

See [search_query.rb#sort_search_results](https://github.com/projecthydra-labs/questioning_authority/blob/linked_data/lib/qa/authorities/linked_data/search_query.rb#L97-#L129)
See [sort_search_results rspec tests](https://github.com/projecthydra-labs/questioning_authority/blob/linked_data/spec/lib/authorities/linked_data/search_query_spec.rb#L201-L271)

elrayle:
I removed the ability to merge a partial config into a full config.  I just don't have time to work through all the questions.  I think it is a useful feature, so I'll create a branch with the merge code that could be added at a later date when I have more time.  It is not a critical feature as there is an alternative approach for apps to modify provided configurations.

elrayle:
@jcoyne, @no-reply, @mbarnett All your requested changes have been addressed.  Please review.

elrayle:
@jcoyne, @no-reply, @mbarnett All your requested changes have been addressed.  Please review.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -0,0 +1,136 @@
+# This module has the primary QA find method.  It also includes methods to process the linked data result and convert
+# it into the expected QA json term result format.
+module Qa::Authorities
+  module LinkedData
+    module FindTerm
```
jcoyne: Could this be extracted to a class TermFinderService perhaps?

lib/qa/authorities/linked_data/config/search_config.rb
jcoyne: Could this be a class rather than a mixin?

lib/qa/authorities/linked_data/rdf_helper.rb
```diff
@@ -0,0 +1,82 @@
+require 'rdf'
+
+# Encapsulates processing of RDF results returned by the linked data authority.  This is used exclussively by Qa::Authorities::LinkedData::GenericAuthority
+# @see Qa::Authorities::LinkedData::GenericAuthority
+module Qa::Authorities
+  module LinkedData
+    module RdfHelper
```
jcoyne: How about not using this as a mixin? That would improve the encapsulation you're talking about above.  Can you make the methods module methods and public (e.g.: `RdfHelper.get_linked_data(url)`)? In this way we know that this module has no dependency on GenericAuthority.

lib/qa/authorities/linked_data/rdf_helper.rb
```diff
@@ -0,0 +1,82 @@
...
+      private
+
+        def get_linked_data(url)
```
jcoyne: `get_` is not ideomatic ruby.  How about changing to `graph_for(url)`

mjgiarlo:
This PR will be reviewed by a group at LDCX or the Hydra Developer Congress next week (March 27-31, 2017).

jeremyf:
ping @no-reply

lib/qa/authorities/linked_data/rdf_helper.rb
hackmastera: this seems like a good suggestion but I know @elrayle is ready to get this merged. Maybe an issue targeting this as a refactor task?

qa.gemspec
```diff
@@ -22,10 +22,13 @@ Gem::Specification.new do |s|
   s.add_dependency "nokogiri", "~> 1.6"
   s.add_dependency "activerecord-import"
   s.add_dependency "deprecation"
+  s.add_dependency "rdf", "2.2.3"
```
hackmastera: This seems like a problem; my sufia 7.3 app currently has 2.2.5 which means I won't be able to upgrade to the next release of qa

lib/qa/authorities/linked_data/rdf_helper.rb
jcoyne: @hackmastera that was an old comment and is no longer relevant after we worked on it at the dev congress.

qa.gemspec
jcoyne: Yes, Gems should never pin dependencies tightly.  Good catch @hackmastera

elrayle: This was supposed to be a temporary pin.  So glad you caught it.  There are two issues with 2.2.5.
1) rspec mocks fail - There is a fix for this.
2) rdf/rdf tries to read file contents with the wrong reader.  (https://github.com/ruby-rdf/rdf/issues/353)  This causes an empty graph to be returned.

elrayle: NOTE: 2.2.5 is not marked as the latest official release in rdf/rdf  2.2.3 is.  But it is what is returned by rubygems.org when you bundle install/update.

README.md
```diff
@@ -333,6 +361,260 @@ This may take a few minutes to finish.
...
+ 
+```
+gem 'rdf-rdfxml'
+```
+
+#### Configuring a LOD Authority
+
+Access to LOD authorities can be configured.  Currently, a configuration exists in QA for OCLC Fast Linked Data, Library of 
+Congress (terms only), and Agrovoc.  Look for configuration files in 
+[/config/authorities/linked_data](https://github.com/ld4l-labs/questioning_authority/tree/linked_data/config/authorities/linked_data).
```
elrayle: RE: Left some inline comments.
Can you point me to those?

RE: Squash
I squashed down to 3 commits.  There were some changes that I would like to be able to track.

RE: README link - changed to tree/master

Not sure why you wouldn't be able to test with Sufia 7.3.  We used this branch with Sufia 7.3 and it worked fine.

README.md
```diff
@@ -333,6 +361,260 @@ This may take a few minutes to finish.
...
+```
+
+NOTES:
+* term: (optional) is used to define how to request term information from the authority and how to interpret results.
+  * url: (required) templated link representation of the authority API URL and mapping of parameters for requesting term information from the authority
+    * template: is the authority API URL with placeholders for substitution parameters in the form {?var_name}
+      * NOTE: {?term_id} (required) and {?subauth} (optional) are expected to match to QA params (see qa_replacement_patterns to match QA params with mapping variables)
+      * Additional substitutions can be made in the authority API if supported by the authority by adding additional mappings.  Search has an example with maximumRecords.
+        * variable: should match a replacement pattern in the template  (e.g. variable: maximumRecords  ==>  {?maximumRecords}
+        * required: true | false  (NOTE: Not enforced at this time.)
```
elrayle: I'd like to open an issue for this after the PR is merged.  There are a few system required substitutions that are checked for.  If a config adds its own substitution and marks it required, there is no check for that.  The required marker is part of the templated URLs spec which is why it is there.  I agree that if it is there, it should be checked.

lib/qa/authorities/linked_data/rdf_helper.rb
```diff
@@ -0,0 +1,102 @@
...
+            raise RDF::FormatError, "Unknown RDF format of results returned by #{uri}. (RDF::FormatError)  You may need to include gem 'linkeddata'."
+          when '404'
+            raise Qa::TermNotFound, "#{uri} Not Found - Term may not exist at LOD Authority. (HTTPNotFound - 404)"
+          when '500'
+            raise Qa::ServiceUnavailable, "#{uri.hostname} on port #{uri.port} is not responding.  Try again later. (HTTPServerError - 500)"
+          when '503'
+            raise Qa::ServiceUnavailable, "#{uri.hostname} on port #{uri.port} is not responding.  Try again later. (HTTPServiceUnavailable - 503)"
+          else
+            raise Qa::ServiceUnavailable, "Unknown error for #{uri.hostname} on port #{uri.port}.  Try again later. (Cause - #{e.message})"
+          end
```
elrayle: I don't believe QA handles timeouts for any authorities, linked data or otherwise.  It might be nice to add that kind of configuration to QA, but it should probably QA wide and not linked data specific.  An issue can be added for this after the PR is merged.

qa.gemspec
jcoyne: I'm not sure I understand that.  It's 2.2.5 here: https://github.com/ruby-rdf/rdf/blob/develop/VERSION

mjgiarlo:
@no-reply Thanks for your review! I am dismissing it now, per the discussion on Hydra Tech last week, so I can merge this PR.

PR #132, opened 2017-02-13
Added support for the crossref REST API - api.crossref.org

geekscruff:
Added support for the crossref REST API - api.crossref.org

Support added for funder and journals only. There are also apis for works, members, prefixes and types.

I can't see anything on the crossref site that prevents use of the public apis in this way.

For funder:
- returns the id rather than the uri
- adds uri into the results
- constructs label and value where label is preflabel + altlabel + location (eg. National Heart, Lung, and Blood Institute, (NHLBI), United States) and value is just the preflabel (National Heart, Lung, and Blood Institute)

```
{"id":"100000050","uri":"http://dx.doi.org/10.13039/100000050","label":"National Heart, Lung, and Blood Institute, (NHLBI), United States","value":"National Heart, Lung, and Blood Institute"}
```

For journals:
- uses issn as id
- adds publisher into the results
- adds issn into results (sometimes there are two ISSNs)

```
{"id":"0730-2347","label":"Texas Heart Institute Journal","publisher":"Texas Heart Institute Journal","issn":["0730-2347","1526-6702"]},
```

Request for feedback. 
- The 'find' method returns the raw json from crossref rather than doing any processing, eg. into the same structure as the results. Is that OK?


mjgiarlo:
Inclined to review this, @jcoyne @awead?

lib/qa/authorities/crossref/generic_authority.rb
```diff
@@ -0,0 +1,59 @@
...
+    def find_url(id)
+      "http://api.crossref.org/#{subauthority}/#{id}"
+    end
+
+    private
+
+      # Reformats the data received from the service
+      def parse_authority_response(response)
+        response['message']['items'].map do |result|
+          case @subauthority
```
awead: shouldn't need the `@` here, I would think.

lib/qa/authorities/crossref/generic_authority.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def initialize(subauthority)
+      @subauthority = subauthority
+    end
+
+    self.label = lambda do |item|
+      [item['name'],
+       item['alt-names'].blank? ? nil : "(#{item['alt-names'].join(', ')})",
+       item['location']].compact.join(', ')
+    end
```
awead: This looks clever, but I'm confused by it. I don't use lambdas very much. It looks like returning an array with a single hash in it. Perhaps some YARD documentation here to explain that. Also, why this way and not just another method call?

geekscruff: I based it on the existing code in geonames (and was learning how to use lambdas) ... I've added some documentation but could easily change it to a method if you think that would be more consistent / understandable?

awead: @geekscruff no, that looks great... except, we use YARD docs. Sorry! Here's a reference https://gist.github.com/chetan/1827484

geekscruff: Sorry, my fault. I've changed the format of the comments.

awead:
@geekscruff thanks for sticking it out. I've approved it and this can be merged as soon as Travis is green.

Issue #133, opened 2017-02-13
LOC response always returns the same JSON 


little9:
I've tested qa 0.11.0 on Rails 4 & 5 and whenever I do an LOC request it always returns this JSON: 

Here's the response for http://localhost:3000/qa/search/loc/subjects?q=China 

```json
[{"id":"info:lc/authorities/subjects/sh2010112128","label":"School sports--Fiction"},{"id":"info:lc/authorities/subjects/sh2008001841","label":"Private flying--Accidents--United States--Periodicals"},{"id":"info:lc/authorities/subjects/sh93000006","label":"Costa Rica--History--1986-"},{"id":"info:lc/authorities/subjects/sh2008116817","label":"Zimbabwe--Economic policy"},{"id":"info:lc/authorities/subjects/sh00008433","label":"Reinstatement"},{"id":"info:lc/authorities/subjects/sh85113459","label":"Revolutionary literature, Portuguese"},{"id":"info:lc/authorities/subjects/sh2003009896","label":"Herbicides--Physiological effect"},{"id":"info:lc/authorities/subjects/sh2003006908","label":"Hôtel de ville (Tours, France)"},{"id":"info:lc/authorities/subjects/sh2010010320","label":"Skin--Cancer--Risk factors--Congresses"},{"id":"info:lc/authorities/subjects/sh85095974","label":"Osteopathic students"},{"id":"info:lc/authorities/subjects/sh2007005866","label":"Face--Muscles--Abnormalities--Treatment"},{"id":"info:lc/authorities/subjects/sh85062079","label":"Horror in art"},{"id":"info:lc/authorities/subjects/sh85008524","label":"Tailed frog"},{"id":"info:lc/authorities/subjects/sh97009005","label":"Apocryphal books (New Testament)--Criticism, interpretation, etc"},{"id":"info:lc/authorities/subjects/sh85011651","label":"Banks and banking, Cooperative"},{"id":"info:lc/authorities/subjects/sh85058833","label":"Harbor masthttp://id.loc.gov/search/?q=China&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fsubjects&format=json
[...]

PR #134, opened 2017-02-13
Faraday params fix

little9:
A fix for issue #133

If you don't have Faraday  use the FlatParamsEncoder it won't use a duplicate param in a request:

?q=china&q=something will become
?q=&q=something

### DIFF
--- a/lib/qa/authorities/web_service_base.rb
+++ b/lib/qa/authorities/web_service_base.rb
@@ -32,6 +32,7 @@ module Qa::Authorities
     # @param url [String]
     # @return [Faraday::Response]
     def response(url)
+      Faraday::Utils.default_params_encoder = Faraday::FlatParamsEncoder
       Faraday.get(url) { |req| req.headers['Accept'] = 'application/json' }
     end
   end


lib/qa/authorities/web_service_base.rb
```diff
@@ -32,6 +32,7 @@ def get_json(url)
     # @param url [String]
     # @return [Faraday::Response]
     def response(url)
+      Faraday::Utils.default_params_encoder = Faraday::FlatParamsEncoder
```
jcoyne: Isn't it already using the `FlatParamsEncoder` in the response method in `Loc::GenericAuthority`?  I'm concerned this would change the behavior for other users of WebServiceBase that do not want `FlatParamsEncoder`

little9: Yeah, that patch does fix it. Sorry!  https://github.com/projecthydra-labs/questioning_authority/blob/43cba871eda61b55096321e10e486254762f939e/lib/qa/authorities/loc/generic_authority.rb

jcoyne:
@little9 maybe you'd want to cut a release of 0.11.1 then?

PR #136, opened 2017-02-20
linked_data branch refactored

elrayle:
The linked_data branch needed significant refactoring.  I know of at least one application that is making use of the linked_data branch.  So I did the refactoring in a separate branch to avoid causing others difficulties.

The refactor includes…
* splitting code into multiple classes that focus on configuration, searching, and term fetching
* created a separate class to process configuration with modules focused on term configuration and search configuration; created complete set of methods for accessing all components of the configuration
* separated all rdf processing into rdf_helper.rb
* added significantly more test coverage


README.md
```diff
@@ -337,9 +365,17 @@ This may take a few minutes to finish.
 
 ### Linked Open Data (LOD) Authorities
 
+You will need to add gems that process the type of linked data returned for the authorities you use.  A gem that covers 
+all formats is [ruby-rdf/linkeddata](https://github.com/ruby-rdf/linkeddata).  This gem is included in QA for development 
+and testing of QA, but is not automatically included in the released gem.  Additionally, it is unlikely that you will need 
+all the formats included by that gem.  You may want to select only those gems that are for the formats you need supported.  
+See all gems in [linkeddata.gemspec](https://github.com/ruby-rdf/linkeddata/blob/develop/linkeddata.gemspec).
```
scande3: While I know to add "gem 'linkeddata'" to my gemfile, spelling out the step explicitly may help someone newer to Rails who just wants to get it up and running. Furthermore, the more advanced configuration explanation might be better in its own wiki page since even I don't know exactly what set of gems I need from linkeddata for this to work and more guidelines of "include X,Y,Z as a base and then A for turtle, B for json-ld, etc" would be useful.

lib/qa/authorities/linked_data/config.rb
```diff
@@ -0,0 +1,77 @@
...
+        @authority_config
+      end
+
+      private
+
+        # Removes characters from the query string that are not tolerated by the API
+        #   See oclc sample code at
+        #   http://experimental.worldcat.org/fast/assignfast/js/assignFASTComplete.js
+        def clean_query_string(q)
+          URI.escape(q.gsub(/-|\(|\)|:/, ""))
```
scande3: Should the gsub be done without the users knowledge? Would it perhaps be better to raise an error if it matched one of those restricted characters so the user can fix their query to this API? Or maybe I am misunderstanding the method and this isn't related to the users query string.

lib/qa/authorities/linked_data/rdf_helper.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+
+        def ioerror_code(e)
+          msg = e.message
+          a = msg.size - 4
+          z = msg.size - 2
+          msg[a..z]
+        end
+
+        def filter_language(graph, language)
```
scande3: Since this looks like the language filter is done on the client side rather than the remote, how hard would it be to allow it to accept a list of languages rather than a single one? Not a feature needed for right now but I often have cases that I need matches in one a set of potential English-style languages.

lib/qa/authorities/linked_data/config.rb
elrayle: I originally copied this from https://github.com/projecthydra-labs/questioning_authority/blob/4cfef296f8ebce5e9c11a350756732c3a8ac3b27/lib/qa/authorities/assign_fast/generic_authority.rb#L52-L57 which was copied from the OCLC js.  I didn't give it much thought.  The exclusions seemed reasonable.  

Now I did a little testing to see what happens.  If I include a `:` or `-` in the query, I get the exact same results.  (In my small test.  It wasn't extensive.)  If I include `(` or `)`, the query fails to process. cURL throws error `-bash: syntax error near unexpected token )`.

Bottom line, I think I agree that this should be removed.  @no-reply, any thoughts?

lib/qa/authorities/linked_data/rdf_helper.rb
elrayle: Shouldn't be too hard.

README.md
elrayle: I can see that.

PR #139, opened 2017-03-17
update tables_generator references qa_local_authority

revgum:
fixes #130 

Ran into this causing me problems with a fresh Hyrax 1.0.0 alpha app, and did some testing to address the references line in the migration. Here's some console output to see what it did:

```
.internal_test_app master % rails g model qa/bleet_bleet_entries qa_local_authority:references label:string uri:string:uniq
[WARNING] The model name 'qa/bleet_bleet_entries' was recognized as a plural, using the singular 'qa/bleet_bleet_entry' instead. Override with --force-plural or setup custom inflection rules for this noun before running the generator.
      invoke  active_record
      create    db/migrate/20170317185708_create_qa_bleet_bleet_entries.rb
      create    app/models/qa/bleet_bleet_entry.rb
   identical    app/models/qa.rb
      invoke    test_unit
      create      test/models/qa/bleet_bleet_entry_test.rb
      create      test/fixtures/qa/bleet_bleet_entries.yml
.internal_test_app master % cat db/migrate/20170317185708_create_qa_bleet_bleet_entries.rb
class CreateQaBleetBleetEntries < ActiveRecord::Migration[5.0]
  def change
    create_table :qa_bleet_bleet_entries do |t|
      t.references :qa_local_authority, foreign_key: true
      t.string :label
      t.string :uri

      t.timestamps
    end
    add_index :qa_bleet_bleet_entries, :uri, unique: true
  end
end

.internal_test_app master % bundle exec rake db:migrate
== 20170317185708 CreateQaBleetBleetEntries: migrating ========================
-- create_table(:qa_bleet_bleet_entries)
   -> 0.0039s
-- add_index(:qa_bleet_bleet_entries, :uri, {:unique=>true})
   -> 0.0008s
== 20170317185708 CreateQaBleetBleetEntries: migrated (0.0048s) ===============

```

PR #141, opened 2017-03-21
Make local tables generator point at the namespaced table for local authorities

### LINKED ISSUE #130: Table generator does invalid substitution.
I see this generated in the `CreateQaLocalAuthorityEntries` migration:
`      t.references :local_authority, foreign_key: true`

but this is the first argument of the gsub in the tables_generator:
 `'t.references :local_authority, index: true, foreign_key: true'`

I don't know why it's trying to match that `index: true`. Perhaps that's a rails 4 thing or a mysql only thing, but we should see if we can do the substitution in a way that matches all cases.

Reported by @acozine

mjgiarlo:
Fixes #130

Supersedes #139.

### DIFF
--- a/lib/generators/qa/local/tables/mysql/mysql_generator.rb
+++ b/lib/generators/qa/local/tables/mysql/mysql_generator.rb
@@ -18,8 +18,8 @@ module Qa::Local
                             .reject { |name| !name.include?('create_qa_local_authority_entries') }.first
         migration_file = File.join('db/migrate', migration_file)
         gsub_file migration_file,
-                  't.references :local_authority, index: true, foreign_key: true',
-                  't.integer :local_authority_id, index: true'
+                  /t\.references :local_authority.*/,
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@revgum Can you point an installation at this branch to see if it fixes the problem you're seeing?

jcoyne:
@mjgiarlo in the future can you keep the rubocop stuff in a separate commit so we can see where the real changes are?

jcoyne:
@mjgiarlo rubocop still has issues.

mjgiarlo:
Rubocop is being weird -- showing me errors on Travis that I don't see locally. Ideally these would not be part of this branch.

mjgiarlo:
@jcoyne I can't get this rubocop violation to trigger locally. Not sure wtf is going on.

mjgiarlo:
@jcoyne OK, this is :green_heart:, finally. I'd like @revgum to confirm this fixes what he was seeing before we merge this, though.

mjgiarlo:
@revgum says it works. Merging.

PR #142, opened 2017-03-30
Extract classes from modules

jcoyne:


lib/qa/authorities/linked_data/config.rb
```diff
@@ -28,8 +28,13 @@ def initialize(auth_name)
...
       end
 
-      include Qa::Authorities::LinkedData::TermConfig
-      include Qa::Authorities::LinkedData::SearchConfig
+      def search
+        @search ||= Qa::Authorities::LinkedData::SearchConfig.new(auth_config.fetch(:search))
+      end
+
+      def term
+        @search ||= Qa::Authorities::LinkedData::TermConfig.new(auth_config.fetch(:term))
```
elrayle: for def term, @search should be @term

lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -4,38 +4,38 @@
...
       def supports_term?
-        !term_config.nil?
+        term_config.present?
       end
 
       # Return term url encoding defined in the configuration for this authority.
       # @return [Hash] the configured term url
       def term_url
-        config_value(term_config, :url)
+        Config.config_value(term_config, :url)
```
elrayle: Search was switched to use fetch instead of config_value.  To be consistent, term should be updated to use hash fetch too.

elrayle:
I tested this with our apps that use the linked_data branch with this branch and they are working.

PR #143, opened 2017-04-29
Add geonames linked data access and usage notes

elrayle:
Supports search query and term fetch for geonames authority.

NOTE: You must get a username from geonames.org.  See [README](https://github.com/projecthydra/questioning_authority/blob/master/README.md#geonames-linked-data) notes for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ You should question your authorities.
       * [Configuring a LOD Authority](#configuring-a-lod-authority)
       * [Query](#query)
       * [Find term](#find-term)
+      * [Authority Usage Notes](#linked-data-authority-usage-notes)
+        * [Agrovoc](#agrovoc-linked-data)
+        * [Geonames](#geonames-linked-data)
+        * [Library of Congress](#library-of-congress-linked-data)
+        * [OCLC FAST](#oclc-fast-linked-data)
       * [Add javascript to support autocomplete](#add-javascript-to-support-autocomplete)
   * [Developer Notes](#developer-notes)
     * [Compatibility](#compatibility)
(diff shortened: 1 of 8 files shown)


jcoyne:
Does this replace https://github.com/projecthydra/questioning_authority/blob/4cfef296f8ebce5e9c11a350756732c3a8ac3b27/lib/qa/authorities/geonames.rb ?

elrayle:
@jcoyne geonames.rb gets results back from geonames as json.  The configuration file added here gets results back as linked data.  The json version does not include the URI, but does include the geonames' id which can be used to construct a URI.

jcoyne:
@elrayle can you explain what you mean by "getting it back as linked data". Are you modeling the results in RDF?

README.md
```diff
@@ -609,6 +614,170 @@ Returns results in the format...
...
+##### Geonames Linked Data
+
+Configuration File: [geonames.json](https://github.com/projecthydra/questioning_authority/blob/master/config/authorities/linked_data/geonames.json)
+
+*Example Search URLs*
+
+Pattern:  `http://www.your.host/qa/search/linked_data/geonames?q={?query}&maximumRecords={?maximumRecords}&username={?username}`
+
+Localhost Examples:
+<br>http://localhost:3000/qa/search/linked_data/geonames?q=ithaca&username=yourusername
```
elrayle: You can set the username's default value, which I should probably mark as recommended.  Since we dumped the merging code, this means that the entire config file has to be copied to the app and changes made to the copy.

There is a comment in the README at line 673 that states how to do this.
https://github.com/samvera/questioning_authority/blob/geonames/README.md#geonames-linked-data
Look at the note after the param description table.

elrayle:
The data that comes back from the authority is a serialization of RDF.  It is normalized to a structure that is the same as those created from non-RDF serializations, except that it also includes a URI.

jcoyne:
@elrayle So, the search functionality is identical and the fetch functionality expects a CGI escaped URI rather than the identifier?

jcoyne:
@elrayle I'm a bit concerned that getting search results back as linked data destroys the relevancy as they aren't returned as a rdf:List:

```
RDF::RDFXML::Reader.open('http://api.geonames.org/search?q=australia&username=demo&type=rdf').each_statement.first
=> #<RDF::Statement:0x3fe94e50086c(<http://sws.geonames.org/2077456/> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.geonames.org/ontology#Feature> .)>
```

elrayle:
@jcoyne I understand the issue with the linked data results not returned as rdf:list or some other indicator of order.  

We are exploring caching authority data, wrapping with a lucene/solr index for search, and providing search results with a predicate that will hold search ranking that can be used for sorting data. 

This isn't in place just yet, but will be in the near future.  This same basic process will be applied across a number of authorities.  We are currently using cached LoC data for our initial implementation test.  Geonames is also being targeted for use with this process.

elrayle:
Instead of creating new authority configuration files in QA itself, the config files are being created in [ld4l-labs/linked_data_authorities](https://github.com/ld4l-labs/linked_data_authorities).

PR #150, opened 2017-06-23
adds active flag to terms defaulting to true

### LINKED ISSUE #149: `active` isn't surfaced in #all, causing unnecessary #find calls
https://github.com/samvera/questioning_authority/blob/master/lib/qa/authorities/local/file_based_authority.rb#L17

If we surface the `active` flag, this would allow the caller to handle selecting the terms much more efficiently. Possibly defaulting to `true` if active doesn't exist in the yml file. 

```
res[:active] = true if res[:active].nil?
{ id: res[:id], label: res[:term], active: res[:active] }.with_indifferent_access
```

This would make the additional `#active?` in Hyrax unnecessary and speed up handling large local authorities. https://github.com/samvera/hyrax/blob/master/app/services/hyrax/qa_select_service.rb#L30

revgum:
fixes #149

### DIFF
--- a/lib/qa/authorities/local/file_based_authority.rb
+++ b/lib/qa/authorities/local/file_based_authority.rb
@@ -14,7 +14,7 @@ module Qa::Authorities
 
     def all
       terms.map do |res|
-        { id: res[:id], label: res[:term] }.with_indifferent_access
+        { id: res[:id], label: res[:term], active: res.fetch(:active, true) }.with_indifferent_access
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/qa/authorities/local/file_based_authority.rb
```diff
@@ -14,7 +14,8 @@ def search(q)
 
     def all
       terms.map do |res|
-        { id: res[:id], label: res[:term] }.with_indifferent_access
+        res[:active] = true if res[:active].nil?
+        { id: res[:id], label: res[:term], active: res[:active] }.with_indifferent_access
```
jcoyne: Can we get rid of line 17 and do `active: res.fetch(:active, true)`?

revgum: @jcoyne oh yea.. of course I had to space that one out. #noob

mjgiarlo:
@revgum I reconfigured Coveralls for this repo such that it only fails when coverage drops 0.3% or more, btw. Should pick that up on the next build.

mjgiarlo:
@jcoyne re-review?

PR #157, opened 2017-12-22
Put the mesh task in the qa namespace

jcoyne:


elrayle:
@jcoyne What is the status of this PR?  It currently has conflicts with master and fails tests.

PR #159, opened 2018-02-08
support the common forms used to define a term template

elrayle:
Supports term templates in the form…
```
"{?term_uri}"              # term_uri is the entire template
"{?term_uri}.rdf"          # term_uri starts the template
"http://localhost/test_default/term?uri={?term_uri}"   # term_uri is a parameter of the template
"http://localhost/test_default/term/{?term_id}"        # term_id is a parameter of the template
```

Fixed errors:
* term_uri needs to be url encoded when it is a parameter
* term_uri should not be encoded when it is the entire template
* substitution of term id was only done when the template parameter was named term_id and not when it was named term_uri, but now happens for both

### DIFF
--- a/lib/qa/authorities/linked_data/config/term_config.rb
+++ b/lib/qa/authorities/linked_data/config/term_config.rb
@@ -112,10 +112,24 @@ module Qa::Authorities
         Config.config_value(term_config, :qa_replacement_patterns)
       end
 
+      # Should the replacement pattern be encoded?
+      # @return [Boolean] true, if the pattern should be encoded; otherwise, false
+      def term_qa_replacement_encoded?(pattern_key)
+        map_key = term_qa_replacement_patterns[pattern_key].to_sym
+        term_replacement_encoded? map_key
+      end
+
       # Are there replacement parameters configured for term fetch?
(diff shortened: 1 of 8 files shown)


lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -176,7 +176,9 @@ def term_subauthority_replacement_pattern
       def term_url_with_replacements(id, sub_auth = nil, replacements = {})
         return nil unless supports_term?
         sub_auth = sub_auth.to_sym if sub_auth.is_a? String
-        url = Config.replace_pattern(term_url_template, term_qa_replacement_patterns[:term_id], id)
+        # do not encode if term parameter is starts the template (e.g. template="{?term_uri}" OR template="{?term_uri}.rdf")
```
cam156: @elrayle I'm wondering if this choice would be better left to the user?  Aka passing a encode parameter into the method.  Is there a reason the user would not know if they want to encode?

elrayle: @cam156 That's good feedback.  I will look and see if I can make that happen in the configuration.

PR #167, opened 2018-07-13
Address dependency vulnerability for rubocop by updating to use bixby

elrayle:


qa.gemspec
```diff
@@ -33,6 +32,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "simplecov"
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "webmock"
-  s.add_development_dependency 'rubocop', '~> 0.42.0'
-  s.add_development_dependency 'rubocop-rspec', '~> 1.8.0'
+  # the hyrax style guide is based on `bixby`. see `.rubocop.yml`
```
RudyOnRails: Should this be first in order to stick with alphabetical ordering?

elrayle: good catch, I'll make that change

elrayle: Looks like that file didn't have any dependencies in alpha order.  I went ahead and  updated them all to be alpha order.

PR #184, opened 2018-11-29
Move rdf_helper methods to rdf_service

elrayle:
A major refactoring of the linked data processing code is underway.  The first step in the process is to extract out `include` code from find_term and search_query processing into separate isolated services.  Work will be committed in multiple PRs to facilitate the review process and avoiding a single massive commit.

This is PR creates the first extracted service.  It collects together only methods that process rdf in the new rdf_service.  These were originally included through the rdf_helper module.

There are a few methods remaining in rdf_helper that will be resolved in a later refactor PR.  The remaining ones will either go away or belong in another place.

PR #185, opened 2018-11-29
Move rdf_helper methods to graph_service

elrayle:
If acceptable, this replaces PR #184.  It is taking a slightly different approach on the replacement service based on feedback.

-----------------

A major refactoring of the linked data processing code is underway. The first step in the process is to extract out include code from find_term and search_query processing into separate isolated services. Work will be committed in multiple PRs to facilitate the review process and avoiding a single massive commit.

This is PR creates the first extracted service. It collects together only methods that process rdf in the new rdf_service. These were originally included through the rdf_helper module.

There are a few methods remaining in rdf_helper that will be resolved in a later refactor PR. The remaining ones will either go away or belong in another place.

### DIFF
--- a/app/services/qa/linked_data/graph_service.rb
+++ b/app/services/qa/linked_data/graph_service.rb
@@ -1,0 +1,77 @@
+# Extend the RDF graph to include additional processing methods.
+module Qa
+  module LinkedData
+    class GraphService
+      attr_reader :graph
+
+      # Retrieve linked data from specified url
+      # @param [String] url from which to retrieve linked data
+      # @param [String | Symbol | Array<String|Symbol>] language for filtering graph (e.g. "en" or :en or ["en", "fr"] or [:en, :fr])
+      # @returns [RDF::Graph] graph of linked data
+      def initialize(url:)
+        @graph = RDF::Graph.load(url)
+      rescue IOError => e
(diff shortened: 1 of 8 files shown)


app/services/qa/linked_data/graph_service.rb
```diff
@@ -0,0 +1,78 @@
...
+    class GraphService
+      attr_reader :graph
+
+      # Retrieve linked data from specified url
+      # @param [String] url from which to retrieve linked data
+      # @param [String | Symbol | Array<String|Symbol>] language for filtering graph (e.g. "en" or :en or ["en", "fr"] or [:en, :fr])
+      # @returns [RDF::Graph] graph of linked data
+      def initialize(url:, language: nil, remove_blanknode_subjects: false)
+        @graph = RDF::Graph.load(url)
+        filter(language: language, remove_blanknode_subjects: remove_blanknode_subjects)
```
cam156: @elrayle I would either remove the filter here, or remove the filter method.  I think I would prefer removing it here, but I could go either way.  If you remove it here you could have the same graph filtered many ways.  I think having both is a bit confusing...

elrayle: I removed filter from the initializer.

PR #186, opened 2018-11-30
Move AuthorityService to app/services

elrayle:
There are no code changes.  The class name of the service did change...

**FROM:**  Qa::Authorities::LinkedData::AuthorityService
**TO:**  Qa::LinkedData::AuthorityService


app/services/qa/linked_data/authority_service.rb
```diff
@@ -32,7 +31,7 @@ def self.authority_configs
 
       # Get the configuration for an authority
       # @param [String] name of the authority
-      # @return [Array<String>] configuration for the specified authority
+      # @return [Hash] configuration for the specified authority
```
elrayle: I caught this copy/paste error while debugging.

PR #187, opened 2018-11-30
Create models for IRI templates and use IRI template service to build URLs

elrayle:
Working toward better separation between the classes.  The IRI related models and the IRI service that builds the URLs have good separation.  The AuthorityUrlService was put in place as a go between for the authority configuration access and the IRI models and service.  More refactoring will happen later to clean up some of the bleeding of config knowledge into the AuthorityUrlService.

### DIFF
--- a/app/models/qa/iri_template/url_config.rb
+++ b/app/models/qa/iri_template/url_config.rb
@@ -1,0 +1,56 @@
+# Provide access to iri template configuration.
+module Qa
+  module IriTemplate
+    class UrlConfig
+      TYPE = "IriTemplate".freeze
+      CONTEXT = "http://www.w3.org/ns/hydra/context.jsonld".freeze
+      attr_reader :template # [String] the URL template with variables for substitution (required)
+      attr_reader :variable_representation # [String] always "BasicRepresentation" # TODO what other values are supported and what do they mean
+      attr_reader :mapping # [Array<Qa::IriTempalte::VariableMap>] array of maps for use with a template (required)
+
+      # @param [Hash] url_template configuration hash for the iri template
+      # @option url_template [String] :template the URL template with variables for substitution (required)
+      # @option url_template [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
+      # @option url_template [Array<Hash>] :mapping array of maps for use with a template (required)
+      def initialize(url_config)
+        @template = extract_template(config: url_config)
+        @mapping = extract_mapping(config: url_config)
+        @variable_representation = url_config.fetch(:variable_representation, 'BasicRepresentation')
+      end
+
+      # Selective extract substitution variable-value pairs from the provided substitutions.
+      # @param [Hash, ActionController::Parameters] full set of passed in substitution values
+      # @returns [HashWithIndifferentAccess] Only variable-value pairs for variables defined in the variable mapping.
+      def extract_substitutions(substitutions)
+        selected_substitutions = HashWithIndifferentAccess.new
+        mapping.each do |m|
+          selected_substitutions[m.variable] = substitutions[m.variable] if substitutions.key? m.variable
+        end
+        selected_substitutions
+      end
+
+      private
+
+        # Extract the url template from the config
+        # @param config [Hash] configuration (json) holding the template to be extracted
+        # @param var [Symbol] key identifying the template in the configuration
+        # @return [String] url template for accessing the authority
+        def extract_template(config:, var: :template)
+          template = config.fetch(var, nil)
+          raise Qa::InvalidConfiguration, "template is required" unless template
+          template
+        end
+
+        # Initialize the variable maps
+        # @param config [Hash] configuration (json) holding the variable maps to be extracted
+        # @param var [Symbol] key identifying the variable mapping array in the configuration
+        # @return [Array<IriTemplate::Map>] array of the variable maps
+        def extract_mapping(config:, var: :mapping)
+          mapping = config.fetch(var, nil)
+          raise Qa::InvalidConfiguration, "mapping is required" unless mapping
+          raise Qa::InvalidConfiguration, "mapping must include at least one map" if mapping.empty?
+          mapping.collect { |m| Qa::IriTemplate::VariableMap.new(m) }
+        end
+    end
+  end
+end
--- a/app/models/qa/iri_template/url_config.rb
+++ b/app/models/qa/iri_template/url_config.rb
@@ -1,4 +1,6 @@
 # Provide access to iri template configuration.
+# See https://www.hydra-cg.com/spec/latest/core/#templated-links for information on IRI Templated Links.
+# TODO: It would be good to find a more complete resource describing templated links.
 module Qa
   module IriTemplate
     class UrlConfig
@@ -8,10 +10,10 @@ module Qa
       attr_reader :variable_representation # [String] always "BasicRepresentation" # TODO what other values are supported and what do they mean
       attr_reader :mapping # [Array<Qa::IriTempalte::VariableMap>] array of maps for use with a template (required)
 
-      # @param [Hash] url_template configuration hash for the iri template
-      # @option url_template [String] :template the URL template with variables for substitution (required)
-      # @option url_template [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
-      # @option url_template [Array<Hash>] :mapping array of maps for use with a template (required)
+      # @param [Hash] url_config configuration hash for the iri template
+      # @option url_config [String] :template the URL template with variables for substitution (required)
+      # @option url_config [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
+      # @option url_config [Array<Hash>] :mapping array of maps for use with a template (required)
       def initialize(url_config)
         @template = extract_template(config: url_config)
         @mapping = extract_mapping(config: url_config)
@@ -32,21 +34,19 @@ module Qa
       private
 
         # Extract the url template from the config
-        # @param config [Hash] configuration (json) holding the template to be extracted
-        # @param var [Symbol] key identifying the template in the configuration
+        # @param config [Hash] configuration holding the template to be extracted
         # @return [String] url template for accessing the authority
-        def extract_template(config:, var: :template)
-          template = config.fetch(var, nil)
+        def extract_template(config:)
+          template = config.fetch(:template, nil)
           raise Qa::InvalidConfiguration, "template is required" unless template
           template
         end
 
         # Initialize the variable maps
-        # @param config [Hash] configuration (json) holding the variable maps to be extracted
-        # @param var [Symbol] key identifying the variable mapping array in the configuration
+        # @param config [Hash] configuration holding the variable maps to be extracted
         # @return [Array<IriTemplate::Map>] array of the variable maps
-        def extract_mapping(config:, var: :mapping)
-          mapping = config.fetch(var, nil)
+        def extract_mapping(config:)
+          mapping = config.fetch(:mapping, nil)
           raise Qa::InvalidConfiguration, "mapping is required" unless mapping
           raise Qa::InvalidConfiguration, "mapping must include at least one map" if mapping.empty?
           mapping.collect { |m| Qa::IriTemplate::VariableMap.new(m) }
--- a/app/models/qa/iri_template/variable_map.rb
+++ b/app/models/qa/iri_template/variable_map.rb
@@ -1,0 +1,79 @@
+# Provide access to iri template variable map configuration.
+module Qa
+  module IriTemplate
+    class VariableMap
+      TYPE = "IriTemplateMapping".freeze
+      attr_reader :variable
+      attr_reader :default
+
+      # @param [Hash] map configuration hash for the variable map
+      # @option map [String] :variable name of the variable in the template (e.g. {?query} has the name 'query')
+      # @option map [String] :property always "hydra:freetextQuery" # TODO what other values are supported and what do they mean
+      # @option map [True | False] :required is this variable required
+      # @option map [String] :default value to use if a value is not provided in the request
+      def initialize(map)
+        @variable = extract_variable(config: map)
+        @required = extract_required(config: map)
+        @default = extract_default(config: map)
+        @property = map.fetch(:property, 'hydra:freetextQuery')
+      end
+
+      # Is this variable required?
+      # @returns true if required; otherwise, false
+      def required?
+        @required
+      end
+
+      # TODO: When implementing more complex query substitution, simple_value is used when template url specifies variable as {var_name}.
+      # Value to use in substitution, using default if one isn't passed in
+      # @param [Object] value to use if it exists
+      # @returns the value to use (e.g. 'fr')
+      def simple_value(sub_value = nil)
+        return sub_value.to_s if sub_value.present?
+        return default if default.present?
+        raise Qa::IriTemplate::MissingParameter, "#{variable} is required, but missing" if required?
+        ''
+      end
+
+      # TODO: When implementing more complex query substitution, parameter_value is used when template url specifies variable as {?var_name}.
+      # # Parameter and value to use in substitution, using default is one isn't passed in
+      # # @param [Object] value to use if it exists
+      # # @returns the parameter and value to use (e.g. 'language=fr')
+      # def parameter_value(sub_value = nil)
+      #   simple_value = simple_value(sub_value)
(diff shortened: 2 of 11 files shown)


app/models/qa/iri_template/url_config.rb
```diff
@@ -0,0 +1,56 @@
...
+  module IriTemplate
+    class UrlConfig
+      TYPE = "IriTemplate".freeze
+      CONTEXT = "http://www.w3.org/ns/hydra/context.jsonld".freeze
+      attr_reader :template # [String] the URL template with variables for substitution (required)
+      attr_reader :variable_representation # [String] always "BasicRepresentation" # TODO what other values are supported and what do they mean
+      attr_reader :mapping # [Array<Qa::IriTempalte::VariableMap>] array of maps for use with a template (required)
+
+      # @param [Hash] url_template configuration hash for the iri template
+      # @option url_template [String] :template the URL template with variables for substitution (required)
```
cam156: @elrayle is this url_config?

app/models/qa/iri_template/url_config.rb
```diff
@@ -0,0 +1,56 @@
...
+      # @option url_template [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
```
cam156: Or is this supposed varriable_representation?

app/models/qa/iri_template/url_config.rb
```diff
@@ -0,0 +1,56 @@
...
+        selected_substitutions
+      end
+
+      private
+
+        # Extract the url template from the config
+        # @param config [Hash] configuration (json) holding the template to be extracted
+        # @param var [Symbol] key identifying the template in the configuration
+        # @return [String] url template for accessing the authority
+        def extract_template(config:, var: :template)
```
cam156: @elrayle since this is private and the second variable is not used, you may want to omit it.  I can see the desire to make it flexible, but it would be simpler not to have it.

If you want to make it easier to add in the future you could just set the variable at the top of the routine.  If this is something you are eyeing for another refactor, just leave it in.

app/models/qa/iri_template/url_config.rb
```diff
@@ -0,0 +1,56 @@
...
+          template = config.fetch(var, nil)
+          raise Qa::InvalidConfiguration, "template is required" unless template
+          template
+        end
+
+        # Initialize the variable maps
+        # @param config [Hash] configuration (json) holding the variable maps to be extracted
+        # @param var [Symbol] key identifying the variable mapping array in the configuration
+        # @return [Array<IriTemplate::Map>] array of the variable maps
+        def extract_mapping(config:, var: :mapping)
```
cam156: Similar to above.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -0,0 +1,79 @@
+# Provide access to iri template variable map configuration.
+module Qa
+  module IriTemplate
+    class VariableMap
+      TYPE = "IriTemplateMapping".freeze
+      attr_reader :variable
```
cam156: @elrayle I liked the comments in the class above in this spot.  I do not mind them below either.  For your future self I would make the place consistent

app/models/qa/iri_template/variable_map.rb
```diff
@@ -0,0 +1,79 @@
...
+        @property = map.fetch(:property, 'hydra:freetextQuery')
+      end
+
+      # Is this variable required?
+      # @returns true if required; otherwise, false
+      def required?
+        @required
+      end
+
+      # TODO: When implementing more complex query substitution, simple_value is used when template url specifies variable as {var_name}.
```
cam156: I'm trying to understand what these two routines mean.  In the larger context does the MappingValue or the context tell how the output should be?  Aka when I construct a VariableMap does it always get evaluated the same way or does the evaluation depend change as it is called.  Things in the construction like property make me feel like it is defined when you instantiate, but these methods make me feel like it is the users choice.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -0,0 +1,79 @@
...
+      #   param_value = "#{variable}=#{simple_value}"
+      # end
+
+      private
+
+        # Extract the variable name from the config
+        # @param config [Hash] configuration (json) holding the variable map
+        # @param var [Symbol] key identifying the variable in the configuration
+        # @return [String] variable for substitution in the url tmeplate
+        def extract_variable(config:, var: :variable)
```
cam156: Similar comment about the second parameter here adding complexity.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -0,0 +1,79 @@
...
+          varname = config.fetch(var, nil)
+          raise Qa::InvalidConfiguration, 'variable is required' unless varname
+          varname
+        end
+
+        # Extract the variable name from the config
+        # @param config [Hash] configuration (json) holding the variable map
+        # @param var [Symbol] key in the configuration identifying whether the variable is required
+        # @return [True | False] required as true or false
+        def extract_required(config:, var: :required)
```
cam156: Similar comment about the second parameter here adding complexity.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -0,0 +1,79 @@
...
+          required = config.fetch(var, nil)
+          raise Qa::InvalidConfiguration, 'required must be true or false' unless required == true || required == false
+          required
+        end
+
+        # Extract the default value from the config ignoring defaults for required variables
+        # @param config [Hash] configuration (json) holding the variable map
+        # @param var [Symbol] key identifying the default value in the configuration
+        # @return [String] default value to use for the variable; nil if variable is required
```
cam156: This comment does not match the code.  Nil is not returned only empty string.  You are taking care of the nil in simple value, although I like the idea of taking care of it down here.  It would simplify simple value.

app/services/qa/iri_template_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      #     * support ExplicitRepresentation
+      #        * literal encoding for values (e.g. value="A simple string" becomes %22A%20simple%20string%22)
+      #        * language encoding for values (e.g. value="A simple string" becomes value="A simple string"@en which is encoded as %22A%20simple%20string%22%40en)
+      #        * type encoding for values (e.g. value=5.5 becomes value="5.5"^^http://www.w3.org/2001/XMLSchema#decimal which is encoded
+      #                                         as %225.5%22%5E%5Ehttp%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23decimal)
+      # Fuller implementations parse the template into component parts and then build the URL by adding parts in as applicable.
+      url = url_config.template
+      url_config.mapping.each do |m|
+        key = m.variable
+        url = url.gsub("{?#{key}}", m.simple_value(substitutions[key])) # Incorrectly applies pattern {?var_name} to produce substitution 'value'
```
cam156: Should this be a "{#{key}}" without the ?

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -0,0 +1,62 @@
...
+    class AuthorityUrlService
+      # Build a url for an authority/subauthority for the specified action.
+      # @param authority [Symbol] name of a registered authority
+      # @param subauthority [String] name of a subauthority
+      # @param action [Symbol] action with valid values :search or :term
+      # @param action_request [String] the request the user is making of the authority (e.g. query text or term id/uri)
+      # @param substitutions [Hash] variable-value pairs to substitute into the URL template
+      # @returns a valid URL the submits the action request to the external authority
+      def self.build_url(action_config:, action:, action_request:, substitutions: {}, subauthority: nil)
+        # auth_config = Qa::Authorities::LinkedData::Config.new(authority)
```
cam156: Should this comment be removed?

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        action_validation(action)
+        # action_config = action_config(auth_config, action)
```
cam156: remove?

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        url_config = Qa::IriTemplate::UrlConfig.new(action_url(action_config, action))
+        selected_substitutions = url_config.extract_substitutions(substitutions)
+        selected_substitutions[action_request_variable(action_config, action)] = action_request
+        selected_substitutions[action_subauth_variable(action_config, action)] = action_subauth_variable_value(action_config, subauthority, action) if subauthority.present?
+
+        Qa::IriTemplateService.build_url(url_config: url_config, substitutions: selected_substitutions)
+      end
+
+      def self.action_validation(action)
```
cam156: I see you putting in this format in again.  One other way to leave the public method static is to make a private initializer and then utilize that in the static public method.  `action` seems to be your obsession this time.

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        return if [:search, :term].include? action
+        raise Qa::UnsupportedAction, "#{action} Not Supported - Action must be one of the supported actions (e.g. :term, :search)"
+      end
+      private_class_method :action_validation
```
cam156: I think it is these lines that I find so hard to read.  It looks like there are a bunch of public class methods on the surface until you dig in.

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      # TODO: elr - rename search and term config methods to be the same to avoid all the ternary checks
```
cam156: You could also have two classes based on the action, one for search and one for ?other?  Then you could instantiate one class or the other based on the action type.

spec/models/iri_template/url_config_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+        },
+        {
+          "@type" => "IriTemplateMapping",
+          "variable" => "param2",
+          "property" => "hydra:freetextQuery",
+          "required" => false,
+          "default" => "echo"
+        }
+      ]
+    }.with_indifferent_access
```
cam156: Do you need the with_indifferent_access here?

app/models/qa/iri_template/variable_map.rb
elrayle: My intention is for all class/module level doc to be the first line of the file.  I'll look to see if I put it elsewhere in another file.

app/models/qa/iri_template/variable_map.rb
elrayle: I'm going to rethink this.  It is defined by the configured mapping.  I agree that having two methods pushes that knowledge off onto the caller even though we should already know whether it is simple or parameter based on the configuration.

elrayle: Ok, here is the problem for the IriTemplate::VariableMap class.  It does not know whether a variable is defined in the Template as simple (e.g. `{var_name}`) or parameter (e.g. `{?var_name}`).  The template determines that...

```
      "template": "http://id.loc.gov/authorities/{subauth}?{?uri}",
```
* `{subauth}` is simple and returns value only
* `{?uri}` is parameter and returns parameter=value

IriTemplate::UrlConfig holds the template and the array of mappings.

IriTemplateService pulls them together to construct the values.  So the way it is written in this PR (albeit commented out for now), the service checks for patterns...

* pattern `{key}` calls simple_value on the variable map
* pattern `{?key}` calls parameter_value on the variable map

In the original implementation, I did this incorrectly.  I haven't been able to find good documentation on IRI Templates, and did some reading between the lines that was incorrect.  I've since found more information (still not great) that indicated the difference in value and parameter substitutions.  In the original implementation, I used `{?key}` as a value substitution only.

In current configurations, templates are configured incorrectly to always use `{?key}`.  So following the incorrect approach, the template example given above would be configured as...

```
      "template": "http://id.loc.gov/authorities/{?subauth}?uri={?uri}",
```
* `{?subauth}` returns value only
* `{?uri}` returns value only  (thus the template has `uri={?uri}`

I am hoping to correct this as part of the refactor process.

app/services/qa/iri_template_service.rb
elrayle: See longer explanation above.  For this to work in the current implementation, it has to follow substitutions for `{?key}` for all mapped variables.  Let me know if the previous explanation doesn't make the reason why clear.

I hope to fix this in a later refactor.

app/models/qa/iri_template/url_config.rb
elrayle: This was an error.  On the @param, it should be the parameter's name (i.e. url_config), then each of the @option lines should also start with the parameter's name.  For each option, after the type, it should have the Hash key name.

https://www.rubydoc.info/gems/yard/file/docs/Tags.md#option

app/models/qa/iri_template/url_config.rb
elrayle: The IRI Template code doesn't yet use any values other than BasicRepresentation, so although this is listed in the options, it isn't processed at this time.

app/models/qa/iri_template/url_config.rb
elrayle: I agree.  I am taking all these out.  Not sure why I did it that way to begin with.

app/models/qa/iri_template/variable_map.rb
elrayle: Good catch.  I adjusted the processing of the default value in general.  The exception in simple_value method for missing required is thrown before the extract_default method is called since it doesn't apply for required variables.

spec/models/iri_template/url_config_spec.rb
elrayle: Removed uses of .with_indifferent_access

app/models/qa/iri_template/url_config.rb
```diff
@@ -1,4 +1,6 @@
 # Provide access to iri template configuration.
+# See https://www.hydra-cg.com/spec/latest/core/#templated-links for information on IRI Templated Links.
+# TODO: It would be good to find a more complete resource describing templated links.
```
elrayle: Added link to documentation of IRI Templated Links to add any devs that come behind me on what it is.

app/models/qa/iri_template/url_config.rb
```diff
@@ -8,10 +10,10 @@ class UrlConfig
...
       attr_reader :mapping # [Array<Qa::IriTempalte::VariableMap>] array of maps for use with a template (required)
 
-      # @param [Hash] url_template configuration hash for the iri template
-      # @option url_template [String] :template the URL template with variables for substitution (required)
-      # @option url_template [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
-      # @option url_template [Array<Hash>] :mapping array of maps for use with a template (required)
+      # @param [Hash] url_config configuration hash for the iri template
+      # @option url_config [String] :template the URL template with variables for substitution (required)
+      # @option url_config [String] :variable_representation always "BasicRepresentation" # TODO what other values are supported and what do they mean
+      # @option url_config [Array<Hash>] :mapping array of maps for use with a template (required)
```
elrayle: Fixes error in yard doc for the Hash options.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -30,9 +32,8 @@ def required?
       # @returns the value to use (e.g. 'fr')
       def simple_value(sub_value = nil)
         return sub_value.to_s if sub_value.present?
-        return default if default.present?
         raise Qa::IriTemplate::MissingParameter, "#{variable} is required, but missing" if required?
-        ''
+        default
```
elrayle: It is more 'right' to throw the exception for required variables before using the default.  Required variables do not have defaults.

app/models/qa/iri_template/url_config.rb
```diff
@@ -0,0 +1,56 @@
+# Provide access to iri template configuration.
+# See https://www.hydra-cg.com/spec/latest/core/#templated-links for information on IRI Templated Links.
+# TODO: It would be good to find a more complete resource describing templated links.
+module Qa
+  module IriTemplate
+    class UrlConfig
+      TYPE = "IriTemplate".freeze
+      CONTEXT = "http://www.w3.org/ns/hydra/context.jsonld".freeze
+      attr_reader :template # [String] the URL template with variables for substitution (required)
```
cam156: @elrayle here is the other place I was talking about with the comments.

PR #188, opened 2018-12-04
Convert GraphService to all class methods

elrayle:
This makes graph service stateless.  A graph can be passed to any of the methods.  It does not matter if the RDF::Graph was created by this service or by some other means.

### DIFF
--- a/lib/qa/authorities/linked_data/search_query.rb
+++ b/lib/qa/authorities/linked_data/search_query.rb
@@ -10,7 +10,7 @@ module Qa::Authorities
         @search_config = search_config
       end
 
-      attr_reader :search_config
+      attr_reader :search_config, :graph
 
       delegate :subauthority?, :supports_sort?, to: :search_config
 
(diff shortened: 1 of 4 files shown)


lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -29,19 +29,18 @@ def search(query, language: nil, replacements: {}, subauth: nil)
...
       end
 
       private
 
         def load_graph(url:, language:)
-          graph_service = Qa::LinkedData::GraphService.new(url: url)
-          graph_service.filter(language: language, remove_blanknode_subjects: true)
-          graph_service.graph
+          @graph = Qa::LinkedData::GraphService.load_graph(url: url)
+          @graph = Qa::LinkedData::GraphService.filter(graph: @graph, language: language) unless language.blank?
```
cam156: I think this will introduce a bug, becuase you are no longer filtering blank nodes

cam156: Looks like you might be missing a test

elrayle:
The failure to filter out blank nodes actually is a bug.  Fixed in PR #190 which also adds a test that would have revealed the bug.

PR #189, opened 2018-12-06
Refactor sort process into a service

elrayle:
The sort process is complex.  The tests for deep_sort_service lays out the various potential sorting scenarios.  Precedence on sort...

* numeric sort, if both are integers
* simple alpha sort of one term, if both are single value
* same language list comparison, if all terms in a list are the same language or no language
* preferred language comparison only, if both lists have the at least one term in the preferred language
* calculate the comparator value for every language individually and sum to get a relative comparison of two multi-language lists (Best I could come up with.  I am open to suggestions.  In most cases, if not all, one of the earlier comparators will be selected.  This is the junk drawer of multi-lingual sorting.)

### DIFF
--- a/app/services/qa/linked_data/deep_sort_service.rb
+++ b/app/services/qa/linked_data/deep_sort_service.rb
@@ -1,0 +1,239 @@
+# Provide service for for sorting an array of hash based on the values at a specified key in the hash.
+module Qa
+  module LinkedData
+    class DeepSortService
+      # @params [Array<Hash<Symbol,Array<RDF::Literal>>>] the array of hashes to sort
+      # @params [sort_key] the key in the hash on whose value the array will be sorted
+      # @returns instance of this class
+      # @example the_array parameter
+      #   [
+      #     {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/names/n2010043281>],
+      #      :id=>[#<RDF::Literal:0x3fcff4a367b4("n 2010043281")>],
+      #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Valli, Sabrina"@en)>],
+      #      :altlabel=>[],
+      #      :sort=>[#<RDF::Literal:0x3fcff54b4c18("2")>]},
+      #     {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/names/n201002344>],
+      #      :id=>[#<RDF::Literal:0x3fcff4a367b4("n 201002344")>],
+      #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Cornell, Joseph"@en)>],
+      #      :altlabel=>[],
+      #      :sort=>[#<RDF::Literal:0x3fcff54b4c18("1")>]}
+      #   ]
+      def initialize(the_array, sort_key, preferred_language = :en)
+        @sortable_elements = the_array.map { |element| DeepSortElement.new(element, sort_key, preferred_language) }
+      end
+
+      # Sort an array of hash on the specified sort key.  The value in the hash at sort key is expected to be an array
+      # with one or more values that are RDF::Literals that translate to a number (e.g. 2), a string number (e.g. "3"),
+      # a string (e.g. "hello"), or a language qualified string (e.g. "hello"@en).
+      # The sort occurs in the following precedence.
+      # * preference for numeric sort (if only one value each and both are integers or a string that can be converted to an integer)
+      # * single value sort (if only one value each and at least one is not an integer)
+      # * multiple values sort (if either has multiple values)
+      # @returns the sorted array
+      # @example returned sorted array
+      #   [
+      #     {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/names/n201002344>],
+      #      :id=>[#<RDF::Literal:0x3fcff4a367b4("n 201002344")>],
+      #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Cornell, Joseph"@en)>],
+      #      :altlabel=>[],
+      #      :sort=>[#<RDF::Literal:0x3fcff54b4c18("1")>]},
+      #     {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/names/n2010043281>],
+      #      :id=>[#<RDF::Literal:0x3fcff4a367b4("n 2010043281")>],
+      #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Valli, Sabrina"@en)>],
+      #      :altlabel=>[],
+      #      :sort=>[#<RDF::Literal:0x3fcff54b4c18("2")>]}
+      #   ]
+      def sort
+        @sortable_elements.sort.map(&:element)
+      end
+
+      class DeepSortElement
+        attr_reader :element, :literals, :preferred_language
+        private :preferred_language
+
+        delegate :size, to: :@literals
+
+        def initialize(element, sort_key, preferred_language)
+          element[sort_key] = Qa::LinkedData::LanguageSortService.new(element[sort_key], preferred_language).sort
+          @element = element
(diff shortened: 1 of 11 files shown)


app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,265 @@
...
+      # a string (e.g. "hello"), or a language qualified string (e.g. "hello"@en).
+      # The sort occurs in the following precedence.
+      # * preference for numeric sort (if only one value each and both are integers or a string that can be converted to an integer)
+      # * single value sort (if only one value each and at least one is not an integer)
+      # * multiple values sort (if either has multiple values)
+      # @returns the sorted array
+      def sort
+        return @sorted_array if @sorted_array.present?
+        presort_sortkey_values
+        @sorted_array = @the_array.sort do |a, b|
```
cam156: I wonder if this would be easier to read if the array was first mapped to a deep sort element and then sorted.  The setup comparison method is a little magic which makes it a smidge hard to read. It may also be more efficient since you are constructing the elements only once instead of at each comparison.
```
   sortable_elements = @the_array.map {|item|  DeepSortElement.new(item[sort_key)}
  @sorted_array = soratble_elements.sort do |a_element, b_element|
      if a_element.integer? && b_element.integer?
          numeric_comparator(a_element, b_element)
     elsif a_element.single? && b_element.single?
          single_value_comparator(a_element, b_element)
     else 
          multiple_value_comparator(a_element, b_element)
     end
  end
```
Or what could be even better is if the DeepSortElement knew how to compare itself with another element and this would read
```
   sortable_elements = @the_array.map {|item|  DeepSortElement.new(item[sort_key)}
  @sorted_array = soratble_elements.sort do |a_element, b_element|
       a_element.compare(b_element)
  end
```

You could then put a bunch of this code into the DeepSortElement class.

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,265 @@
...
+        end
+
+        def no_language?(language)
+          language.nil?
+        end
+    end
+  end
+end
+
+class DeepSortElement
```
cam156: To follow my logic from above further the compare method here could look like
```
   def compare(other_element)
     if integer? && other_element.integer?
          numeric_comparator(other_element)
     elsif single? && other_element.single?
          single_value_comparator(other_element)
     else 
          multiple_value_comparator(other_element)
     end
  end

 def numeric_comparator(other_element)
     integer <=> other_element.integer
 end

...
```

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,265 @@
...
+          next single_value_comparator if both_single?
+          multiple_value_comparator
+        end
+      end
+
+      private
+
+        def presort_sortkey_values
+          the_array.map! do |element|
+            element[sort_key] = Qa::LinkedData::LanguageSortService.new(element[sort_key], preferred_language).sort
```
cam156: If the array was mapped to DeepSortElements you could take advantage of sorted_literals in that class here?

app/services/qa/linked_data/deep_sort_service.rb
elrayle: I like that idea.  I'll look at making that adjustment.

randalldfloyd:
@elrayle @cam156 Just throwing in an initial response from a first pass...

Any custom sort is going to be complex so there's no way around that, but I was able to follow the logic and syntax for the most part.  The way it is used and supposed to work is still eluding me a bit so I'm going to read the tests to see it in action and maybe that will help.

I was also trying to think if any of the heavy lifting or setup could be simplified by using the Comparable mixin.  Not sure it would make it better or worse or just not applicable:  

https://docs.ruby-lang.org/en/2.5.0/Comparable.html. 
https://medium.com/@farsi_mehdi/the-comparable-mixin-8a3096fa1a0a
https://rubyplus.com/articles/2791-Ruby-Comparable-Basics

I don't have experience with it in Ruby but I've seen it in Java.  I think the gist of it is that you get to supply behavior for comparison operators that objects are going to get for free anyway (or something like that) which could potentially simplify a layer of comparison setup.

elrayle:
@cam156 @randalldfloyd Reworked DeepSortService to make better use of ruby <=> and comparators patterns.

elrayle:
@randalldfloyd I took a look at the links you provided.  Thanks for adding those. 

RE: Comparable mixin - The doc states "Comparable uses <=> to implement the conventional comparison operators (<, <=, ==, >=, and >) and the method between?."  I don't think we need these extra methods so I didn't add the mixin.

elrayle:
I think I'm going to make one more change to this.  It bothers me that DeepSortElement is a class that anyone can use since it is specific to the DeepSortService.  I'm going to make it a private class using the approach described here...  https://blog.arkency.com/2016/02/private-classes-in-ruby/

elrayle:
Done and ready for review.

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,244 @@
...
+        end
+
+        def language(idx = 0, list = literals)
+          return list[idx].language if list[idx].respond_to?(:language)
+          nil
+        end
+
+        def includes_preferred_language?
+          return @has_preferred_language if @has_preferred_language.present?
+          filtered = filtered_literals(preferred_language)
```
cam156: Given the literals are sorted by language you could just test `language == preferred_language` since the first item will either be in the prefered language or it is not in the list at all.

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,244 @@
...
+          @has_preferred_language = filtered.size.positive?
+        end
+
+        def all_same_language?
+          return @all_same_language if @all_same_language.present?
+          @all_same_language = true
```
cam156: Since the literal list is sorted by language you should be able to compare `language(0) == language(size-1)` to see if they are all the same

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,244 @@
...
+          1.upto(size - 1) { |idx| return @all_same_language = false unless language(idx) == language(0) }
+          @all_same_language
+        end
+
+        def languages
+          filtered_literals_by_language.keys
+        end
+
+        def filtered_literals(filter_language)
+          filtered_literals_by_language.key?(filter_language) ? filtered_literals_by_language[filter_language] : []
```
cam156: this can be simplified into `filtered_literals_by_language.fetch(filter_language,[])` see https://ruby-doc.org/core-2.5.1/Hash.html#method-i-fetch

app/services/qa/linked_data/deep_sort_service.rb
```diff
@@ -0,0 +1,244 @@
...
+
+          def filtered_literals_by_language
+            @filtered_literals_by_language ||= create_all_filters
+          end
+
+          def create_all_filters
+            bins = {}
+            0.upto(size - 1) do |idx|
+              lang = language(idx, literals)
+              filter = bins.key?(lang) ? bins[lang] : []
```
cam156: This can also be simplified by fetch `filter = bins.fetch(lang, [])`

app/services/qa/linked_data/deep_sort_service.rb
elrayle: Done

app/services/qa/linked_data/deep_sort_service.rb
elrayle: Done

app/services/qa/linked_data/deep_sort_service.rb
elrayle: Done

app/services/qa/linked_data/deep_sort_service.rb
elrayle: Done

elrayle:
@cam156 Requested changes are complete.

PR #190, opened 2018-12-07
Add test to catch when blank nodes are not filtered out of search results

elrayle:
* Fixes bug caught in review of PR #188 
* Adds test that would have revealed the bug.

### DIFF
--- a/lib/qa/authorities/linked_data/search_query.rb
+++ b/lib/qa/authorities/linked_data/search_query.rb
@@ -37,7 +37,7 @@ module Qa::Authorities
 
         def load_graph(url:, language:)
           @graph = Qa::LinkedData::GraphService.load_graph(url: url)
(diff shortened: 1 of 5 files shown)


lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -37,7 +37,7 @@ def search(query, language: nil, replacements: {}, subauth: nil)
 
         def load_graph(url:, language:)
           @graph = Qa::LinkedData::GraphService.load_graph(url: url)
-          @graph = Qa::LinkedData::GraphService.filter(graph: @graph, language: language) unless language.blank?
+          @graph = Qa::LinkedData::GraphService.filter(graph: @graph, language: language, remove_blanknode_subjects: true)
```
elrayle: Don't need the `unless language.blank?` check because the GraphService will make that check.

PR #192, opened 2018-12-17
Extract out search results mapping to services

elrayle:
* adds GraphMapperService that selects out a list of predicate-values from an RDF graph for a single URI
* adds a SearchMapperService that uses the GraphMapperSerivce for each search result


lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -109,14 +85,6 @@ def wrap_labels(labels)
           lbl = '[' + lbl + ']' if labels.size > 1
           lbl
         end
-
-        def sort_search_results(json_results)
```
elrayle: sort happens in mapper

app/services/qa/linked_data/mapper/search_results_mapper_service.rb
```diff
@@ -0,0 +1,47 @@
...
+          #      :id=>[#<RDF::Literal:0x3fcff4a367b4("n 2010043281")>],
+          #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Valli, Sabrina"@en)>],
+          #      :altlabel=>[],
+          #      :sort=>[#<RDF::Literal:0x3fcff54b4c18("2")>]}
+          #   ]
+          def map_values(graph:, predicate_map:, sort_key:, preferred_language: nil)
+            search_matches = []
+            graph.subjects.each do |subject|
+              next if subject.anonymous? # skip blank nodes
+              values = Qa::LinkedData::Mapper::GraphMapperService.map_values(graph: graph, predicate_map: predicate_map, subject_uri: subject)
```
randalldfloyd: @elrayle Just curious, in cases like this where classes share the same module namespace, is it necessary to fully qualify the class name?  The tests pass (for me) with just `GraphMapperService` instead of the qualified form - is there a runtime error that could still arise?  Just wondering, those lines get really long with long module names and parameters.

PR #193, opened 2019-01-09
remove obsolete URL processing methods

elrayle:
Ref PR #187 Create models for IRI templates and use IRI template service to build URLs

When IriTemplateService was created, several methods in the config files became obsolete, but didn't get removed as part of that PR.


cam156:
@elrayle this one looks ok, except the coverage went down on your configurations.  You may need to add some tests just for those.  Also Travis is acting strange in a few builds, but I do not think that is isolated to this PR.  I have heard a couple people in the office mention travis and bundler.

elrayle:
Turns out there are a few more methods that are also obsolete causing test coverage to go down.  Nice when the tools help us to find issues.  I'm in the process of removing the additional obsolete methods now.

cam156:
@elrayle looks like replacement_encoded and qa_replacement_encoded are still untested.  Can those be removed too?

PR #195, opened 2019-01-10
add ability to encode the variable value

elrayle:
Encoding is primarily used to encode URIs that are part of the URL path instead of being a parameter.  But the generalized processing allows any variable to be marked as requiring URL encoding.

Also making all methods private unless they are needed in other classes.  In this refactor, that effect #default and #required?

The encoding support was in the original term_url_with_replacements which is obsolete with PR #187.  That functionality should have been pulled into variable maps then.  This PR incorporates this functionality in a broader way that allows any variable to be encoded based on configurations.

### DIFF
--- a/app/models/qa/iri_template/variable_map.rb
+++ b/app/models/qa/iri_template/variable_map.rb
@@ -1,29 +1,30 @@
 # Provide access to iri template variable map configuration.
 # See https://www.hydra-cg.com/spec/latest/core/#templated-links for information on IRI Templated Links - Variable Mapping.
 # TODO: It would be good to find a more complete resource describing templated links.
+require 'erb'
+
 module Qa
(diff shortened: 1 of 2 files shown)


app/models/qa/iri_template/variable_map.rb
```diff
@@ -1,39 +1,42 @@
...
+
 module Qa
   module IriTemplate
     class VariableMap
+      include ERB::Util
+
       TYPE = "IriTemplateMapping".freeze
       attr_reader :variable
-      attr_reader :default
+      attr_reader :default, :required, :encode, :property
```
cam156: I do not think the required attr_reader is used anywhere.  Since it is private it might be better to just remove it.  It looks like you utilize required? which uses @required

PR #197, opened 2019-01-15
add configuration models for additional context

elrayle:
These are the classes to process the context configuration.  They are not being used yet.

Remaining work:
* In search_config, read in and process the context configuration using these classes
* Add context_mapping_service, which will use these configurations to extract the context from the graph and return the results as json.

This PR also defines generalized helper methods for fetching configs from the config hash.  This includes fetching required configs, boolean configs, and configs values that need to be converted to a symbol.

### DIFF
--- a/app/models/qa/iri_template/variable_map.rb
+++ b/app/models/qa/iri_template/variable_map.rb
@@ -13,24 +13,24 @@ module Qa
       attr_reader :default
       private :default
 
-      # @param [Hash] map configuration hash for the variable map
-      # @option map [String] :variable (required) name of the variable in the template (e.g. {?query} has the name 'query')
-      # @option map [String] :property (optional) always "hydra:freetextQuery" # TODO what other values are supported and what do they mean
-      # @option map [Boolean] :required (required) is this variable required
-      # @option map [String] :default (optional) value to use if a value is not provided in the request (default: '')
-      # @option map [Boolean] :encode (optional) whether to url_encode the value (default: false)
-      def initialize(map)
-        @variable = extract_variable(config: map)
-        @required = extract_required(config: map)
-        @default = extract_default(config: map)
-        @encode = extract_encode(config: map)
-        @property = map.fetch(:property, 'hydra:freetextQuery')
+      # @param [Hash] variable_map configuration hash for the variable map
+      # @option variable_map [String] :variable (required) name of the variable in the template (e.g. {?query} has the name 'query')
+      # @option variable_map [String] :property (optional) always "hydra:freetextQuery" # TODO what other values are supported and what do they mean
+      # @option variable_map [Boolean] :required (required) is this variable required
+      # @option variable_map [String] :default (optional) value to use if a value is not provided in the request (default: '')
+      # @option variable_map [Boolean] :encode (optional) whether to url_encode the value (default: false)
+      def initialize(variable_map)
+        @variable = Qa::LinkedData::Config::Helper.fetch_required(variable_map, :variable, nil)
+        @required = Qa::LinkedData::Config::Helper.fetch_boolean(variable_map, :required, nil)
+        @default = Qa::LinkedData::Config::Helper.fetch(variable_map, :default, '').to_s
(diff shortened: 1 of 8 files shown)


app/models/qa/iri_template/variable_map.rb
```diff
@@ -13,23 +13,23 @@ class VariableMap
...
+      # @param [Hash] variable_map configuration hash for the variable map
+      # @option variable_map [String] :variable (required) name of the variable in the template (e.g. {?query} has the name 'query')
+      # @option variable_map [String] :property (optional) always "hydra:freetextQuery" # TODO what other values are supported and what do they mean
+      # @option variable_map [Boolean] :required (required) is this variable required
+      # @option variable_map [String] :default (optional) value to use if a value is not provided in the request (default: '')
+      # @option variable_map [Boolean] :encode (optional) whether to url_encode the value (default: false)
+      def initialize(variable_map)
+        @variable = Qa::LinkedData::Config::Helper.fetch_required(variable_map, :variable, nil)
+        @required = Qa::LinkedData::Config::Helper.fetch_boolean(variable_map, :required, nil)
+        @default = variable_map.fetch(:default, '').to_s
```
cam156: @elrayle No need for changes, but I am noticing an inconstancy here.  In Some cases you use the Helper class and in other cases you access the map directly.  It might be better in the long run to be consistent.

app/models/qa/iri_template/variable_map.rb
```diff
@@ -13,23 +13,23 @@ class VariableMap
...
+        @encode = Qa::LinkedData::Config::Helper.fetch_boolean(variable_map, :encode, false)
+        @property = variable_map.fetch(:property, 'hydra:freetextQuery')
       end
 
       # TODO: When implementing more complex query substitution, simple_value is used when template url specifies variable as {var_name}.
       # Value to use in substitution, using default if one isn't passed in
-      # @param [Object] value to use if it exists
+      # rparam [Object] value to use if it exists
```
cam156: @elrayle this looks like a typo

app/models/qa/linked_data/config/context_map.rb
```diff
@@ -0,0 +1,75 @@
...
+module Qa
+  module LinkedData
+    module Config
+      class ContextMap
+        attr_reader :properties # [Array<Qa::LinkedData::Config::ContextPropertyMap>] set of property map models
+
+        attr_reader :context_map, :groups
+        private :context_map, :groups
+
+        # @param [Hash] context_map that defines groups and properties for additional context to display during the selection process
```
cam156: 👍 Comments look great!

app/models/qa/linked_data/config/context_map.rb
```diff
@@ -0,0 +1,75 @@
...
+        #       }
+        #     ]
+        #   }
+        def initialize(context_map = {})
+          @context_map = context_map
+          extract_groups
+          extract_properties
+        end
+
+        def group_label(group_id)
```
cam156: You are allowing the user to get a list of properties back, but not a list of groups.  They need to know the id to get at group information.  I'm guessing this is how you are using the data, but getting a list of groups seems like it might be nice.

app/models/qa/iri_template/variable_map.rb
elrayle: nice catch

app/models/qa/linked_data/config/context_map.rb
elrayle: I am trying to only surface the public API what is actually being used.  When started out, I was surfacing what I thought would be useful and decided against that to avoid public API bloat.  If someone requests the list of groups, it can easily be added at that time.

app/models/qa/iri_template/variable_map.rb
elrayle: This makes sense.  I'll make that change.

elrayle:
* fixed typo rparam to @param
* fixed typo @returns to @return
* consistently use helper class for all access of the config hash (in new and recently refactored classes only)

PR #200, opened 2019-01-25
process additional context and add to final json results

elrayle:
**Adds:**
* context=true|false parameter to QA search URL
* uses ldpath gem to get context property's values based on ldpath in config for the property
* add context_mapper_service which fetches the context for each result and adds it to the result hash

----
**Remaining Work after this PR:**
* expand values that are URIs to include URI + label (will be in later PR)
* figure out why some ldpaths do not parse (handles this gracefully, so will be in later PR)
* ldpath gem loads (and re-loads) a new graph many times over when the data is available in the passed in graph.  Need to explore performance improvements for ldpath.

### DIFF
--- a/app/controllers/qa/linked_data_terms_controller.rb
+++ b/app/controllers/qa/linked_data_terms_controller.rb
@@ -35,7 +35,7 @@ class Qa::LinkedDataTermsController < ::ApplicationController
   # get "/search/linked_data/:vocab(/:subauthority)"
   # @see Qa::Authorities::LinkedData::SearchQuery#search
   def search
-    terms = @authority.search(query, subauth: subauthority, language: language, replacements: replacement_params)
+    terms = @authority.search(query, subauth: subauthority, language: language, replacements: replacement_params, context: context?)
     cors_allow_origin_header(response)
     render json: terms
   rescue Qa::ServiceUnavailable
@@ -176,7 +176,7 @@ class Qa::LinkedDataTermsController < ::ApplicationController
     end
 
     def subauth_warn_msg
-      subauthority.nil? ? "" : " sub-authority #{subauthority} in"
+      subauthority.blank? ? "" : " sub-authority #{subauthority} in"
     end
 
     def format
@@ -186,7 +186,12 @@ class Qa::LinkedDataTermsController < ::ApplicationController
     end
 
     def jsonld?
-      format == 'jsonld'
+      format.casecmp('jsonld').zero?
+    end
+
+    def context?
+      context = params.fetch(:context, 'false')
+      context.casecmp('true').zero?
     end
 
     def validate_auth_reload_token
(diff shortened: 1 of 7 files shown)


app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -189,6 +189,11 @@ def jsonld?
       format == 'jsonld'
     end
 
+    def context?
+      context = params.fetch(:context, false)
+      context == 'true' ? true : false
```
cam156: @elrayle Do you care about TRUE or True here?

app/services/qa/linked_data/mapper/context_mapper_service.rb
```diff
@@ -0,0 +1,46 @@
...
+        self.graph_service = Qa::LinkedData::GraphService
+
+        class << self
+          # Extract predicates specified in the predicate_map from the graph and return as a value map for a single subject URI.
+          # @param graph [RDF::Graph] the graph from which to extract result values
+          # @param context_map [Qa::LinkedData::Config::ContextMap] defines properties to extract from the graph to provide additional context
+          # @param subject_uri [RDF::URI] the subject within the graph for which the values are being extracted
+          # @return [<Hash<Symbol><Array<Object>>] mapped context values and information with hash of map key = array of object values for predicates identified in predicate_map.
+          # @example value map for a single result
+          #   {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/names/n2010043281>],
```
cam156: This should start with a `[`

cam156: I also thing if I am reading the code correctly the example should really look like below.  I do not see id, label, altlabel being added in construct context.
```
[{"group" => "group label,
   "property" => "property label",
   "values" => ["value 1","value 2"],
  "selectable" => true,
  "drillable" => false}]
```

app/controllers/qa/linked_data_terms_controller.rb
elrayle: Yep... I'll downcase before comparison.  Good catch.

app/services/qa/linked_data/mapper/context_mapper_service.rb
elrayle: That is a copy paste error.  I'll update the example.

app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -186,7 +186,12 @@ def format
     end
 
     def jsonld?
-      format == 'jsonld'
+      format.casecmp('jsonld').zero?
+    end
+
+    def context?
+      context = params.fetch(:context, 'false')
+      context.casecmp('true').zero?
```
elrayle: Switched to use casecmp based on rubocop recommendation.  Applied the same change to jsonld? for consistency.  

And one other small cleanup to check for blank? instead of nil? for subauth, which covers more potential non-existing values for subauth.

elrayle:
@cam156 Your requested changes have been made.

PR #202, opened 2019-02-08
Create ldpath program for context property map

elrayle:
Also moves block processing for graph_mapper_service outside of the individual property to the set of properties.  This allows for adding context to the set of properties returned.

Example:
```
{ 
  id: 1, 
  label: 'The title', 
  context:
  {
    location: '42.4488° N, 76.4763° W'
  }
}
```

### DIFF
--- a/app/services/qa/linked_data/mapper/graph_mapper_service.rb
+++ b/app/services/qa/linked_data/mapper/graph_mapper_service.rb
@@ -29,9 +29,9 @@ module Qa
           value_map = {}
           predicate_map.each do |key, predicate|
             values = predicate == :subject_uri ? [subject_uri] : graph_service.object_values(graph: graph, subject: subject_uri, predicate: predicate)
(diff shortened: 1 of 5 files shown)


app/services/qa/linked_data/mapper/graph_mapper_service.rb
```diff
@@ -29,9 +29,9 @@ def self.map_values(graph:, predicate_map:, subject_uri:)
           value_map = {}
           predicate_map.each do |key, predicate|
             values = predicate == :subject_uri ? [subject_uri] : graph_service.object_values(graph: graph, subject: subject_uri, predicate: predicate)
-            values = yield values if block_given?
             value_map[key] = values
           end
+          value_map = yield value_map if block_given?
```
elrayle: This should have been here to allow appending extra keys to the set instead of trying to effect a single value in the set.  This will be used to append the context to the set.

PR #196, opened 2019-01-11
Fix deprecation

bbarberBPL:
I'm currently refactoring one of our gems and qa v2.2. However when running the tests I kept getting 

```/home/bbarber/.rbenv/versions/2.4.5/lib/ruby/gems/2.4.0/gems/qa-3.0.0/lib/qa/authorities/loc/generic_authority.rb:27:in `build_query_u
rl': warning: URI.escape is obsolete
/home/bbarber/.rbenv/versions/2.4.5/lib/ruby/gems/2.4.0/gems/qa-3.0.0/lib/qa/authorities/loc/generic_authority.rb:28:in `build_query_url'
: warning: URI.escape is obsolete
/home/bbarber/.rbenv/versions/2.4.5/lib/ruby/gems/2.4.0/gems/qa-3.0.0/lib/qa/authorities/loc/generic_authority.rb:27:in `build_query_url'
: warning: URI.escape is obsolete
/home/bbarber/.rbenv/versions/2.4.5/lib/ruby/gems/2.4.0/gems/qa-3.0.0/lib/qa/authorities/loc/generic_authority.rb:28:in `build_query_url'
: warning: URI.escape is obsolete```

I bumped the version up in my `gemspec` to the latest release and found i was getting the same warnings in my tests 

Doing further research online I found on the ruby docs [here](https://ruby-doc.org/stdlib-2.4.0/libdoc/uri/rdoc/URI/Escape.html)  That as of 2.4.0 this method has been deprecated and is now obsolete. Since I am running on 2.4.5 I first was able to make a monkey patch in the gem. But my colleague advised me to inform the Samvera community about this. I also removed a duplicate declaration of `simplecov` that was listed in both the `gemspec` and the `Gemfile`

I was able to set up engine cart and run the specs and all of them passed.  I created a method in the web_service_base module that will maintain backwards compatibility for earlier ruby versions but for anyone using 2.4.0 and above it will use `CGI.escape` instead of `URI.escape`.

PR #210, opened 2019-03-04
Replace URI.escape with ERB::Util.url_encode

elrayle:
This is the closest replacement to what already exists.  There are multiple discussions on what is the correct way to do this.  We may want to revisit later.

See… https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape

This addresses deprecation warnings associated with `URI.escape`

### DIFF
--- a/lib/qa/authorities/assign_fast/generic_authority.rb
+++ b/lib/qa/authorities/assign_fast/generic_authority.rb
@@ -53,7 +53,7 @@ module Qa::Authorities
       #   See oclc sample code at
       #   http://experimental.worldcat.org/fast/assignfast/js/assignFASTComplete.js
       def clean_query_string(q)
-        URI.escape(q.gsub(/-|\(|\)|:/, ""))
+        ERB::Util.url_encode(q.gsub(/-|\(|\)|:/, ""))
       end
 
       def parse_authority_response(raw_response)
--- a/lib/qa/authorities/linked_data/config.rb
+++ b/lib/qa/authorities/linked_data/config.rb
@@ -40,7 +40,7 @@ module Qa::Authorities
       end
 
       def config_version
-        @config_version ||= authority_config.fetch(:QaConfigVersion, '1.0')
+        @config_version ||= authority_config.fetch(:QA_CONFIG_VERSION, '1.0')
       end
 
       def config_version?(version)
--- a/lib/qa/authorities/linked_data/config.rb
+++ b/lib/qa/authorities/linked_data/config.rb
@@ -75,9 +75,11 @@ module Qa::Authorities
           convert_1_0_url_to_2_0_url(:term)
         end
 
+        # @deprecated Update to linked data config version 2.0 instead
         def convert_1_0_url_to_2_0_url(action_key)
           url_template = @authority_config.fetch(action_key, {}).fetch(:url, {}).fetch(:template, "")
           return if url_template.blank?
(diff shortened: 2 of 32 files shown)


lib/qa/authorities/linked_data/config.rb
```diff
@@ -18,10 +17,6 @@
 module Qa::Authorities
   module LinkedData
     class Config
-      class << self
```
elrayle: This should have been removed previously when the use of ERB was moved to another class.

elrayle:
This is related to the work done in PR #196.

elrayle:
Confirmed each modified authority by comparing results before and after for the following URLs...

* http://localhost:3000/qa/search/assign_fast/all?q=periodic+table
* http://localhost:3000/qa/search/crossref/journals?q=chest
* http://localhost:3000/qa/search/geonames?q=port
* http://localhost:3000/qa/search/getty/aat?q=amphora
* http://localhost:3000/qa/search/getty/tgn?q=memphis
* http://localhost:3000/qa/search/getty/ulan?q=washington
* http://localhost:3000/qa/search/loc/subjects?q=History--
* http://localhost:3000/qa/search/linked_data/oclc_fast?q=twain&maximumRecords=10

elrayle:
This replaces PR #196 which made a fix for a single authority.  This makes a consistent fix across all uses of `URI.escape`.

elrayle:
@bbarberBPL As this replaces PR #196, can you review and comment on the changes in this PR?

lib/qa/authorities/assign_fast/generic_authority.rb
```diff
@@ -53,7 +53,7 @@ def build_query_url(q)
       #   See oclc sample code at
       #   http://experimental.worldcat.org/fast/assignfast/js/assignFASTComplete.js
       def clean_query_string(q)
-        URI.escape(q.gsub(/-|\(|\)|:/, ""))
+        ERB::Util.url_encode(q.gsub(/-|\(|\)|:/, ""))
```
boveus: I think `ERB::Util.url_encode` will properly encode all of those except for the dash.  I tested it in IRB really quick
```
2.3.6 :003 > ERB::Util.url_encode("-|\(|\)|:")
 => "-%7C%28%7C%29%7C%3A"
```
I am not really sure  how this is being used elsewhere, so I am not sure what removing the gsub would potentially break.  Allowing the special characters may improve the autosuggest functionality slightly by allowing users to search for terms containing special characters

elrayle:
I tried before and after for...

* http://localhost:3000/qa/search/loc/names?q=Queer%20as%20folk%20(Television%20program%20:%20U.S.)
* http://localhost:3000/qa/search/loc/names?q=King%20Cole%20of%20Spokane,%20Inc.

Both worked in the before and after tests.

elrayle:
@hackmastera You appear to be the original committer for assign_fast.  There is a question about the clean_query_string method for assign_fast.  The call to `URI.escape` is being replaced by `ERB::Util.url_encode`.  The original  code includes a gsub that appears to remove some characters.  With `ERB::Util.url_encode`, these characters appear to be encoded properly.  I am wondering two things...

* Is the new approach adequate and the gsub can be removed?
* Do you have an example that requires the gsub that can be used for before and after testing?

hackmastera:
@elrayle It looks like `assign_fast` is still in use in https://github.com/sciencehistory/chf-sufia/blob/master/app/assets/javascripts/edit_metadata.js; maybe @jrochkind or @eddierubeiz could identify some tests?

My vague memory is that the characters I removed were an issue on the API side, rather than a matter of encoding.

PR #223, opened 2019-04-16
allow language to be specified in the http accept_language header

elrayle:
Adds a few other improvements…
* allow * as a wildcard for language processing which will prevent filtering of literals by language
* do not include parameters that do not have a value when building URLs
* clean template built URL to remove any occurances of ?&, &&, or ending with &

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
           project: qa
 
       - samvera/engine_cart_generate:
-          cache_key: v1-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
+          cache_key: v2-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
 
       - samvera/bundle_for_gem:
           ruby_version: << parameters.ruby_version >>
--- a/app/controllers/qa/linked_data_terms_controller.rb
+++ b/app/controllers/qa/linked_data_terms_controller.rb
@@ -180,7 +180,9 @@ class Qa::LinkedDataTermsController < ::ApplicationController
     end
 
     def language
-      params[:lang]
+      request_language = request.env['HTTP_ACCEPT_LANGUAGE']
+      request_language = request_language.scan(/^[a-z]{2}/).first if request_language.present?
+      params[:lang] || request_language
     end
 
     def subauthority
@@ -188,7 +190,7 @@ class Qa::LinkedDataTermsController < ::ApplicationController
     end
 
     def replacement_params
-      params.reject { |k, _v| ['q', 'vocab', 'controller', 'action', 'subauthority', 'language', 'id'].include?(k) }
+      params.reject { |k, _v| ['q', 'vocab', 'controller', 'action', 'subauthority', 'lang', 'id'].include?(k) }
     end
 
     def subauth_warn_msg
--- a/app/services/qa/iri_template_service.rb
+++ b/app/services/qa/iri_template_service.rb
@@ -1,31 +1,39 @@
 # Provide service for building a URL based on an IRI Templated Link and its variable mappings based on provided substitutions.
 module Qa
   class IriTemplateService
-    # Construct an url from an IriTemplate making identified substitutions
-    # @param url_config [Qa::IriTemplate::UrlConfig] configuration (json) holding the template and variable mappings
-    # @param substitutions [HashWithIndifferentAccess] name-value pairs to substitute into the url template
-    # @return [String] url with substitutions
-    def self.build_url(url_config:, substitutions:)
-      # TODO: This is a very simple approach using direct substitution into the template string.
-      #   Better would be to...
-      #     * patterns without a substitution are not included in the resulting URL
-      #     * appropriately adds '?' or '&'
-      #     * ensure proper escaping of values (e.g. value="A simple string" which is encoded as A%20simple%20string)
-      #   Even more advanced would be to...
-      #     * support BasicRepresentation (which is what it does now)
-      #     * support ExplicitRepresentation
-      #        * literal encoding for values (e.g. value="A simple string" becomes %22A%20simple%20string%22)
-      #        * language encoding for values (e.g. value="A simple string" becomes value="A simple string"@en which is encoded as %22A%20simple%20string%22%40en)
-      #        * type encoding for values (e.g. value=5.5 becomes value="5.5"^^http://www.w3.org/2001/XMLSchema#decimal which is encoded
-      #                                         as %225.5%22%5E%5Ehttp%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23decimal)
-      # Fuller implementations parse the template into component parts and then build the URL by adding parts in as applicable.
-      url = url_config.template
-      url_config.mapping.each do |m|
-        key = m.variable
-        url = url.gsub("{#{key}}", m.simple_value(substitutions[key]))
-        url = url.gsub("{?#{key}}", m.parameter_value(substitutions[key]))
+    class << self
+      # Construct an url from an IriTemplate making identified substitutions
+      # @param url_config [Qa::IriTemplate::UrlConfig] configuration (json) holding the template and variable mappings
+      # @param substitutions [HashWithIndifferentAccess] name-value pairs to substitute into the url template
+      # @return [String] url with substitutions
+      def build_url(url_config:, substitutions:)
+        # TODO: This is a very simple approach using direct substitution into the template string.
+        #   Better would be to...
+        #     * appropriately adds '?' or '&'
+        #     * ensure proper escaping of values (e.g. value="A simple string" which is encoded as A%20simple%20string)
+        #   Even more advanced would be to...
+        #     * support BasicRepresentation (which is what it does now)
+        #     * support ExplicitRepresentation
+        #        * literal encoding for values (e.g. value="A simple string" becomes %22A%20simple%20string%22)
+        #        * language encoding for values (e.g. value="A simple string" becomes value="A simple string"@en which is encoded as %22A%20simple%20string%22%40en)
+        #        * type encoding for values (e.g. value=5.5 becomes value="5.5"^^http://www.w3.org/2001/XMLSchema#decimal which is encoded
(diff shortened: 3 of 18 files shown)


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           project: qa
 
       - samvera/engine_cart_generate:
-          cache_key: v1-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
+          cache_key: v2-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
```
elrayle: Required to force Circle CI cache to rebuild and pick up the new linked_data fixtures.

app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -180,15 +180,17 @@ def id
     end
 
     def language
-      params[:lang]
+      request_language = request.env['HTTP_ACCEPT_LANGUAGE']
+      request_language = request_language.scan(/^[a-z]{2}/).first if request_language.present?
+      params[:lang] || request_language
```
elrayle: Extends language specification to honor HTTP_ACCEPT_LANGUAGE if specified.

app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -180,15 +180,17 @@ def id
...
     end
 
     def subauthority
       params[:subauthority]
     end
 
     def replacement_params
-      params.reject { |k, _v| ['q', 'vocab', 'controller', 'action', 'subauthority', 'language', 'id'].include?(k) }
+      params.reject { |k, _v| ['q', 'vocab', 'controller', 'action', 'subauthority', 'lang', 'id'].include?(k) }
```
elrayle: This should have always been lang instead of language.

app/services/qa/iri_template_service.rb
```diff
@@ -1,31 +1,39 @@
...
-      #        * type encoding for values (e.g. value=5.5 becomes value="5.5"^^http://www.w3.org/2001/XMLSchema#decimal which is encoded
-      #                                         as %225.5%22%5E%5Ehttp%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23decimal)
-      # Fuller implementations parse the template into component parts and then build the URL by adding parts in as applicable.
-      url = url_config.template
-      url_config.mapping.each do |m|
-        key = m.variable
-        url = url.gsub("{#{key}}", m.simple_value(substitutions[key]))
-        url = url.gsub("{?#{key}}", m.parameter_value(substitutions[key]))
+    class << self
+      # Construct an url from an IriTemplate making identified substitutions
```
elrayle: This is mostly an indentation change by adding class << self.  This allows the private methods to be specified under the private section.

app/services/qa/iri_template_service.rb
```diff
@@ -1,31 +1,39 @@
...
+        clean(url)
       end
-      url
+
+      private
+
+        # In the process of substitution, if a value is missing, you can end up with '?&', '&&', or ending with '&'.  These are all removed this method.
+        def clean(url)
+          url.gsub(/&&*/, '&').chomp('&').gsub('?&', '?')
+        end
```
elrayle: The clean method is the only change to this service.

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -24,9 +25,10 @@ def action_validation(action)
             raise Qa::UnsupportedAction, "#{action} Not Supported - Action must be one of the supported actions (e.g. :term, :search)"
           end
 
-          def combined_substitutions(action_config, action, action_request, substitutions, subauthority)
+          def combined_substitutions(action_config, action, action_request, substitutions, subauthority, language) # rubocop:disable Metrics/ParameterLists
             substitutions[action_request_variable(action_config, action)] = action_request
-            substitutions[action_subauth_variable(action_config)] = action_subauth_variable_value(action_config, subauthority) if subauthority.present?
+            substitutions[action_subauth_variable(action_config)] = action_subauth_variable_value(action_config, subauthority) if supports_subauthorities?(action_config) && subauthority.present?
+            substitutions[action_language_variable(action_config)] = language_value(language) if supports_language_parameter?(action_config) && language.present?
```
elrayle: `lang` used to be passed in as part of the `substitutions`. This was an artifact of `lang` failing to be removed in the controller.  Now that it is properly removed, it has to be passed in by name.  This is cleaner as it makes the language processing more visible.

app/services/qa/linked_data/language_service.rb
```diff
@@ -22,6 +24,7 @@ def literal_has_language_marker?(literal)
           def normalize_language(language)
             return language if language.blank?
             language = [language] unless language.is_a? Array
+            return nil if language.include?(WILDCARD)
```
elrayle: By setting the language to nil, there will be no filtering.  When an authority configuration setw the language to `*`, it overrides the site wide default and prevents filtering for that authority.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -37,7 +37,7 @@ def initialize(term_config)
       def find(id, language: nil, replacements: {}, subauth: nil, jsonld: false)
         raise Qa::InvalidLinkedDataAuthority, "Unable to initialize linked data term sub-authority #{subauth}" unless subauth.nil? || term_subauthority?(subauth)
         @language = Qa::LinkedData::LanguageService.preferred_language(user_language: language, authority_language: term_config.term_language)
-        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: id, substitutions: replacements, subauthority: subauth)
+        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: id, substitutions: replacements, subauthority: subauth, language: @language)
```
elrayle: After determining the preferred_language, it was then being ignored by passing `language` to the AuthorityUrlService.  Changing this to `@language` causes the preferred language to be correctly passed instead.

lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -35,7 +35,7 @@ def search(query, language: nil, replacements: {}, subauth: nil, context: false)
         raise Qa::InvalidLinkedDataAuthority, "Unable to initialize linked data search sub-authority #{subauth}" unless subauth.nil? || subauthority?(subauth)
         @context = context
         @language = language_service.preferred_language(user_language: language, authority_language: search_config.language)
-        url = authority_service.build_url(action_config: search_config, action: :search, action_request: query, substitutions: replacements, subauthority: subauth)
+        url = authority_service.build_url(action_config: search_config, action: :search, action_request: query, substitutions: replacements, subauthority: subauth, language: @language)
```
elrayle: Same here.  Needs to pass `@language` which is the preferred langauge, instead of the user/http language which may be nil if not set.

spec/lib/authorities/linked_data/search_config_spec.rb
```diff
@@ -12,7 +13,7 @@
         url: {
           :@context => 'http://www.w3.org/ns/hydra/context.jsonld',
           :@type => 'IriTemplate',
-          template: 'http://localhost/test_default/search?{?subauth}&{?query}&{?param1}&{?param2}',
+          template: 'http://localhost/test_default/search?{?subauth}&{?query}&{?param1}&{?param2}&{?lang}',
```
elrayle: The updates to the expected `full_config`  here corresponds to the changes to the `lod_full_config.json` fixture.

spec/lib/authorities/linked_data/term_config_spec.rb
```diff
@@ -12,7 +13,7 @@
         url: {
           :@context => 'http://www.w3.org/ns/hydra/context.jsonld',
           :@type => 'IriTemplate',
-          template: 'http://localhost/test_default/term/{subauth}/{term_id}?{?param1}&{?param2}',
+          template: 'http://localhost/test_default/term/{subauth}/{term_id}?{?param1}&{?param2}&{?lang}',
```
elrayle: The updates to the expected full_config here corresponds to the changes to the lod_full_config.json fixture.

spec/lib/authorities/linked_data/term_config_spec.rb
```diff
@@ -156,7 +172,7 @@
       expect(min_config.term_language).to eq nil
     end
     it 'returns the preferred language for selecting literal values if configured for term' do
-      expect(full_config.term_language).to eq [:en]
+      expect(full_config.term_language).to eq [:es]
```
elrayle: This value was changed in `lod_full_config.json` fixture to make each language different so we can be sure the language we are checking came for the right place in the config.

spec/services/linked_data/language_service_spec.rb
```diff
@@ -8,33 +8,74 @@
     let(:authority_language) { nil }
 
     context 'when neither user nor authority language are passed in' do
-      it 'returns default language from Qa configuration' do
-        expect(subject).to match_array [:en]
+      context 'and site wide default language is set' do
+        before do
```
elrayle: These test changes are covering the various scenarios for setting language, including the new WILDCARD.

app/services/qa/linked_data/authority_url_service.rb
```diff
@@ -35,13 +37,30 @@ def action_request_variable(action_config, action)
...
+          def supports_language_parameter?(action_config)
+            action_config.supports_language_parameter?
+          end
+
+          def action_language_variable(action_config)
+            action_config.qa_replacement_patterns[:lang]
+          end
+
+          def language_value(language)
+            return nil if language.blank? || !language.is_a?(Array)
```
carolyncole: @elrayle Just checking the logic here.  If the language is 'EN' you will return nil, but if the language is ['EN'] you will return 'EN' . Is this correct?

elrayle: Will return nil when language is...  nil, :en, 'en'
Will return :en when language is... ['en']

The reason this is ok is because the parameter is expected to be an array of symbols (line 12)
```
         # @param language [Array<Symbol>] languages for filtering returned literals (optional) 
```

The language is normalized to be an `Array<Symbol>` by the [language_service](https://github.com/samvera/questioning_authority/blob/5cd213de71a10db79f5914ee6524392ee07a2cbd/app/services/qa/linked_data/language_service.rb#L24-L29).

I could raise an exception if language is not `nil` or an `Array<Symbol>` to make this clearer.

elrayle: Went with removing `|| !language.is_a?(Array)` since the param doc says it should be an array.

PR #225, opened 2019-04-18
extract out ldpath processing into ldpath service

elrayle:
Extract out the ldpath processing into the ldpath service.  This facilitates the next step which will be using ldpath with linked data configuration predicates specified outside the extended context section of the configuration.

This includes the ability to tell ldpath not to look outside the passed in graph.  Otherwise, ldpath may make multiple network calls per single QA request.  In many, if not most, cases, this is not performant.  By default, this is set to true which prevents ldpath from making additional network calls.

### DIFF
--- a/app/models/qa/linked_data/config/context_property_map.rb
+++ b/app/models/qa/linked_data/config/context_property_map.rb
@@ -67,7 +67,7 @@ module Qa
         # Values of this property for a specfic subject URI
         # @return [Array<String>] values for this property
         def values(graph, subject_uri)
-          ldpath_evaluate(basic_program, graph, subject_uri)
+          Qa::LinkedData::LdpathService.ldpath_evaluate(program: basic_program, graph: graph, subject_uri: subject_uri)
         end
 
         # Values of this property for a specfic subject URI with URI values expanded to include id and label.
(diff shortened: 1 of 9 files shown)


app/models/qa/linked_data/config/context_property_map.rb
```diff
@@ -67,7 +67,7 @@ def expand_uri?
         # Values of this property for a specfic subject URI
         # @return [Array<String>] values for this property
         def values(graph, subject_uri)
-          ldpath_evaluate(basic_program, graph, subject_uri)
+          Qa::LinkedData::LdpathService.ldpath_evaluate(program: basic_program, graph: graph, subject_uri: subject_uri)
```
elrayle: switched to the new ldpath service

lib/generators/qa/install/templates/config/initializers/qa.rb
```diff
@@ -14,4 +14,8 @@
   # For linked data access, specify default language for sorting and selection.  The default is only used if a language is not
   # specified in the authority's configuration file and not passed in as a parameter.  (e.g. :en, [:en], or [:en, :fr])
   # config.default_language = :en
+
+  # When true, prevents ldpath requests from making additional network calls.  All values will come from the context graph
+  # passed to the ldpath request.
+  # config.limit_ldpath_to_context = true
```
elrayle: Add ability to determine whether ldpath will look outside the passed in context graph.

lib/qa.rb
```diff
@@ -24,6 +24,12 @@ def self.config(&block)
     @config
   end
 
+  def self.deprecation_warning(in_msg: nil, msg:)
+    return if Rails.env == 'test'
+    in_msg = in_msg.present? ? "In #{in_msg}, " : ''
+    warn "[DEPRECATED] #{in_msg}#{msg}  It will be removed in the next major release."
+  end
```
elrayle: Finer control over deprecation messages.

spec/models/linked_data/config/context_property_map_spec.rb
```diff
@@ -209,34 +209,6 @@
     it 'returns the values selected from the graph' do
       expect(subject.values(graph, subject_uri)).to match_array coordinates
     end
-
-    context 'when ldpath_program gets parse error' do
-      let(:ldpath) { property_map[:ldpath] }
```
elrayle: moved to tests for new ldpath service

app/services/qa/linked_data/ldpath_service.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'ldpath'
+
+module Qa
+  module LinkedData
+    class LdpathService
+      VALUE_ON_ERROR = [].freeze
+
+      # Create the ldpath program for a given ldpath.
+      # @param [String] ldpath to follow to get a value from a graph (documation: http://marmotta.apache.org/ldpath/language.html)
+      # @param [Hash] shortcut names for URI prefixes with key = part of predicate that is the same for all terms (e.g. { "madsrdf": "http://www.loc.gov/mads/rdf/v1#" })
```
carolyncole: You are missing the term `prefixes` here after `[Hash]`

app/services/qa/linked_data/ldpath_service.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      # Evaluate an ldpath for a specific subject uri in the context of a graph and return the extracted values.
+      # @param [Ldpath::Program] an executable program that will extract a value from a graph
+      # @param [RDF::Graph] the graph
+      # @param [RDF::URI] the subject uri
+      ## @return [Array<String>] the extracted values based on the ldpath
+      def self.ldpath_evaluate(program:, graph:, subject_uri:)
+        return VALUE_ON_ERROR if program.blank?
+        output = program.evaluate(subject_uri, context: graph, limit_to_context: Qa.config.limit_ldpath_to_context?)
```
carolyncole: @elrayle This is tying behavior to the configuration in a bit of a hidden way.  It is not horrible, but I might change the method signature to this to surface the connection:
```
def self.ldpath_evaluate(program:, graph:, subject_uri:, limit_to_context: Qa.config.limit_ldpath_to_context?)
  return VALUE_ON_ERROR if program.blank?
  output = program.evaluate(subject_uri, context: graph, limit_to_context: limit_to_context)
```

It also has the advantage of allowing single case overrides if needed at some point (testing could be easier)

lib/qa/configuration.rb
```diff
@@ -37,5 +37,13 @@ def valid_authority_reload_token?(token)
     def default_language
       @default_language ||= :en
     end
+
+    # When true, prevents ldpath requests from making additional network calls.  All values will come from the context graph
+    # passed to the ldpath request.
+    attr_writer :limit_ldpath_to_context
+    def limit_ldpath_to_context?
+      return true if @limit_ldpath_to_context.nil?
```
carolyncole: could be written as, which seems a bit simpler:
```
 @limit_ldpath_to_context ||= true
 @limit_ldpath_to_context
```

app/services/qa/linked_data/ldpath_service.rb
elrayle: Double checked [yardoc documentation](https://www.rubydoc.info/gems/yard/file/docs/Tags.md#param).    For named params, the param name goes before the type.  I'll update the yardoc in this file and others already touched by this PR.

I might need to check other places to see if it is used consistently.  But I'll do that under a separate PR.

lib/qa/configuration.rb
elrayle: That is how I had it originally, but if `@limit_ldpath_to_context == false`, it will see the `||=` check as not being set and change the value to true, thus causing `@limit_ldpath_to_context` to always be true.

app/services/qa/linked_data/ldpath_service.rb
elrayle: Turns out this was the only file effected.

elrayle:
@carolyncole Your requested changes are complete and tests are green.

lib/qa/configuration.rb
carolyncole: @elrayle you are so right!  Sorry!

PR #226, opened 2019-04-18
Update search/term configs to be able to read in results config specified as ldpath

elrayle:
This is also a precursor to extend configurations as ldpath beyond extended context section of the config.


elrayle:
I pushed this with PR #225 code in this branch.

PR #227, opened 2019-04-18
Update search/term configs to be able to read in results config specified as ldpath

elrayle:
This is also a precursor to extend configurations as ldpath beyond extended context section of the config.


elrayle:
Replaces PR #226 which accidentally included code beyond the current PR.

PR #230, opened 2019-04-25
update tests to use real values found in the data

elrayle:
The particular stubbing that is happening when using graph filtering in QA does not seem to mind if bogus data is used.  But as we move to ldpath, it is looking for real data in the graph.

This is prep work for moving findterm to use ldpath.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ jobs:
           project: qa
 
       - samvera/engine_cart_generate:
-          cache_key: v2-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
+          cache_key: v4-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
 
       - samvera/bundle_for_gem:
           ruby_version: << parameters.ruby_version >>
--- a/lib/qa/authorities/linked_data/config.rb
+++ b/lib/qa/authorities/linked_data/config.rb
@@ -32,7 +32,7 @@ module Qa::Authorities
       end
 
       def term
-        @term ||= Qa::Authorities::LinkedData::TermConfig.new(authority_config.fetch(:term), prefixes)
+        @term ||= Qa::Authorities::LinkedData::TermConfig.new(authority_config.fetch(:term), prefixes, self)
       end
 
       def prefixes
--- a/lib/qa/authorities/linked_data/config/term_config.rb
+++ b/lib/qa/authorities/linked_data/config/term_config.rb
@@ -5,16 +5,19 @@
 module Qa::Authorities
   module LinkedData
     class TermConfig
-      attr_reader :prefixes
+      attr_reader :prefixes, :full_config, :term_config
+      private :full_config, :term_config
+
+      delegate :authority_name, to: :full_config
 
       # @param [Hash] config the term portion of the config
-      def initialize(config, prefixes = {})
+      # @param [Hash<Symbol><String>] prefixes URL map of prefixes to use with ldpaths
+      # @param [Qa::Authorities::LinkedData::Config] full_config the full linked data configuration that the passed in search config is part of
+      def initialize(config, prefixes = {}, full_config = nil)
         @term_config = config
         @prefixes = prefixes
+        @full_config = full_config
       end
-
-      attr_reader :term_config
-      private :term_config
 
       # Does this authority configuration have term defined?
       # @return [True|False] true if term fetching is configured; otherwise, false
--- a/lib/qa/authorities/linked_data/find_term.rb
+++ b/lib/qa/authorities/linked_data/find_term.rb
@@ -13,7 +13,7 @@ module Qa::Authorities
       attr_reader :term_config, :full_graph, :filtered_graph, :language
       private :full_graph, :filtered_graph, :language
 
-      delegate :term_subauthority?, to: :term_config
+      delegate :term_subauthority?, :prefixes, :authority_name, to: :term_config
 
       # Find a single term in a linked data authority
       # @param [String] the id of the term to fetch
@@ -37,7 +37,8 @@ module Qa::Authorities
       def find(id, language: nil, replacements: {}, subauth: nil, jsonld: false)
         raise Qa::InvalidLinkedDataAuthority, "Unable to initialize linked data term sub-authority #{subauth}" unless subauth.nil? || term_subauthority?(subauth)
         @language = Qa::LinkedData::LanguageService.preferred_language(user_language: language, authority_language: term_config.term_language)
-        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: id, substitutions: replacements, subauthority: subauth, language: @language)
+        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: normalize_id(id),
+                                                            substitutions: replacements, subauthority: subauth, language: @language)
         Rails.logger.info "QA Linked Data term url: #{url}"
         load_graph(url: url)
         return "{}" unless full_graph.size.positive?
@@ -59,10 +60,19 @@ module Qa::Authorities
           results = extract_preds(filtered_graph, preds_for_term)
           consolidated_results = consolidate_term_results(results)
           json_results = convert_term_to_json(consolidated_results)
-          termhash = select_json_result_for_id(json_results, id)
+          termhash = select_json_result_for_id(json_results, normalize_id(id))
           predicates_hash = predicates_with_subject_uri(termhash[:uri])
           termhash['predicates'] = predicates_hash unless predicates_hash.length <= 0
           termhash
+        end
+
+        def normalize_id(id)
+          return id if expects_uri?
+          authority_name.to_s.casecmp('loc').zero? ? id.delete(' ') : id
+        end
+
+        def expects_uri?
+          term_config.term_id_expects_uri?
         end
 
         def preds_for_term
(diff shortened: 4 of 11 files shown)


spec/controllers/linked_data_terms_controller_spec.rb
```diff
@@ -390,48 +390,48 @@
     context 'producing internal server error' do
       context 'when server returns 500' do
         before do
-          stub_request(:get, 'http://localhost/test_default/term?uri=http://test.org/530369').to_return(status: 500)
+          stub_request(:get, 'http://localhost/test_default/term?uri=http://id.worldcat.org/fast/530369').to_return(status: 500)
```
elrayle: The worldcat uri is what is really in the data.  The test.org uri doesn't exist in the graph.  When working with ldpath, the data needs to be in the graph.

spec/fixtures/authorities/linked_data/lod_lang_defaults.json
```diff
@@ -4,21 +4,21 @@
     "url": {
       "@context": "http://www.w3.org/ns/hydra/context.jsonld",
       "@type":    "IriTemplate",
-      "template": "http://localhost/test_default/term/{term_id}",
+      "template": "{term_uri}",
```
elrayle: This could have been updated to use the workcat request API instead of the bogus `localhost/test_default`, but it is more flexible to simply use `{term_uri}`.  This allows this config to be used with multiple authorities for testing.

spec/fixtures/authorities/linked_data/lod_lang_multi_defaults.json
```diff
@@ -4,21 +4,21 @@
     "url": {
       "@context": "http://www.w3.org/ns/hydra/context.jsonld",
       "@type":    "IriTemplate",
-      "template": "http://localhost/test_default/term/{term_id}",
+      "template": "{term_uri}",
```
elrayle: same

spec/fixtures/authorities/linked_data/lod_term_uri_param_config.json
```diff
@@ -20,7 +20,7 @@
     },
     "term_id": "URI",
     "results": {
-      "id_predicate":       "http://id.loc.gov/vocabulary/identifiers/lccn",
+      "id_predicate":       "http://purl.org/dc/terms/identifier",
```
elrayle: The data actually uses this predicate for the id.

spec/fixtures/lod_loc_term_found.rdf.xml
```diff
@@ -116,7 +116,7 @@
       </madsrdf:Topic>
     </madsrdf:hasVariant>
     <madsrdf:editorialNote>headings beginning with the word [Scientific;] and subdivision [Science] under ethnic groups and individual wars, e.g. [World War, 1939-1945--Science]</madsrdf:editorialNote>
-    <identifiers:lccn xmlns:identifiers="http://id.loc.gov/vocabulary/identifiers/">sh 85118553</identifiers:lccn>
+    <identifiers:lccn xmlns:identifiers="http://id.loc.gov/vocabulary/identifiers/">sh85118553</identifiers:lccn>
```
elrayle: This is an ugly change.  I am in discussions with loc.  For the id predicate, the id includes a blank (e.g. `sh 85118553`), but the uri does not.  The current loc config expects the id to be passed in without a blank so it can be substituted into the URL template.  But the code that tries to identify the URI from the results, looks for the id as the object value and assumes that the URI for the matching triple is the subject_uri being fetched.  In this case, the id in the object value does want the blank.

For now, I am removing the blank from the fixture data file representing loc data.  I believe the long term fix is to force loc config to require the URI to be passed in for find.  Then we don't have to figure out how to translate from id to uri.

My hesitation to go there now is that will cause a major version as it will not be backward compatible for existing loc configs.

The other option is to add kludge code for loc that adds/removes the blank as needed for loc only.  Yuck.

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -128,9 +128,9 @@
             context "and NO language defined in Qa config" do
               let(:lod_lang_no_defaults) { described_class.new(term_config(:LOD_LANG_NO_DEFAULTS)) }
               let :results do
-                stub_request(:get, "http://localhost/test_no_default/term/c_9513")
+                stub_request(:get, "http://aims.fao.org/aos/agrovoc/c_9513")
```
elrayle: Use real agrovoc api url instead of fake `localhost/test_no_default`

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -128,9 +128,9 @@
...
                   .to_return(status: 200, body: webmock_fixture("lod_lang_term_enfr.rdf.xml"), headers: { 'Content-Type' => 'application/rdf+xml' })
-                lod_lang_no_defaults.find('c_9513')
+                lod_lang_no_defaults.find('http://aims.fao.org/aos/agrovoc/c_9513')
```
elrayle: Agrovoc data do not have id triples, so it has to be called with the full URI.

.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           project: qa
 
       - samvera/engine_cart_generate:
-          cache_key: v2-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
+          cache_key: v4-internal-test-app-{{ checksum "qa.gemspec" }}-{{ checksum "spec/test_app_templates/lib/generators/test_app_generator.rb" }}-{{ checksum "lib/generators/qa/install/install_generator.rb" }}-<< parameters.rails_version >>-<< parameters.ruby_version >>
```
elrayle: Forces cache to rebuild the test app which is required due to changes to the fixture authority configs.

lib/qa/authorities/linked_data/config.rb
```diff
@@ -32,7 +32,7 @@ def search
       end
 
       def term
-        @term ||= Qa::Authorities::LinkedData::TermConfig.new(authority_config.fetch(:term), prefixes)
+        @term ||= Qa::Authorities::LinkedData::TermConfig.new(authority_config.fetch(:term), prefixes, self)
```
elrayle: Passes in the full config so term config can delegate methods like `authority_name`.

lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -5,17 +5,20 @@
...
+      attr_reader :prefixes, :full_config, :term_config
+      private :full_config, :term_config
+
+      delegate :authority_name, to: :full_config
 
       # @param [Hash] config the term portion of the config
-      def initialize(config, prefixes = {})
+      # @param [Hash<Symbol><String>] prefixes URL map of prefixes to use with ldpaths
+      # @param [Qa::Authorities::LinkedData::Config] full_config the full linked data configuration that the passed in search config is part of
+      def initialize(config, prefixes = {}, full_config = nil)
```
elrayle: Add link back to the full config so it can be used for delegation.

lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -5,17 +5,20 @@
...
         @term_config = config
         @prefixes = prefixes
+        @full_config = full_config
       end
 
-      attr_reader :term_config
-      private :term_config
-
```
elrayle: Moved up with other attr_reader defs.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -37,7 +37,8 @@ def initialize(term_config)
       def find(id, language: nil, replacements: {}, subauth: nil, jsonld: false)
         raise Qa::InvalidLinkedDataAuthority, "Unable to initialize linked data term sub-authority #{subauth}" unless subauth.nil? || term_subauthority?(subauth)
         @language = Qa::LinkedData::LanguageService.preferred_language(user_language: language, authority_language: term_config.term_language)
-        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: id, substitutions: replacements, subauthority: subauth, language: @language)
+        url = Qa::LinkedData::AuthorityUrlService.build_url(action_config: term_config, action: :term, action_request: normalize_id(id),
```
elrayle: Passes `normalized_id(id)` to handle any special processing that needs to happen to the id before generating the URL.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -59,12 +60,21 @@ def parse_term_authority_response(id)
           results = extract_preds(filtered_graph, preds_for_term)
           consolidated_results = consolidate_term_results(results)
           json_results = convert_term_to_json(consolidated_results)
-          termhash = select_json_result_for_id(json_results, id)
+          termhash = select_json_result_for_id(json_results, normalize_id(id))
```
elrayle: Same normalization needs to happen here.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -59,12 +60,21 @@ def parse_term_authority_response(id)
...
           predicates_hash = predicates_with_subject_uri(termhash[:uri])
           termhash['predicates'] = predicates_hash unless predicates_hash.length <= 0
           termhash
         end
 
+        def normalize_id(id)
+          return id if expects_uri?
+          authority_name.to_s.casecmp('loc').zero? ? id.delete(' ') : id
```
elrayle: This is special processing for LOC.  In the triple, the id is specified with a blank (e.g. `sh 123`).  But to request the entity by URI, the blank cannot exist (e.g. http://id.loc.gov/authorities/subjects/sh123`.  The URI is generated from the passed in id.  If the blank is not removed, an exception is thrown for a malformed URI.

spec/controllers/linked_data_terms_controller_spec.rb
```diff
@@ -378,7 +378,7 @@
             .to_return(status: 200, body: webmock_fixture('lod_loc_term_found.rdf.xml'), headers: { 'Content-Type' => 'application/rdf+xml' })
         end
         it 'succeeds and defaults to json content type' do
-          get :show, params: { id: 'sh85118553', vocab: 'LOC', subauthority: 'subjects' }
+          get :show, params: { id: 'sh 85118553', vocab: 'LOC', subauthority: 'subjects' }
```
elrayle: Use a real LOC id in the request to be sure the blank is removed properly.

PR #232, opened 2019-04-30
Refactor findterm to use term_results_mapper_service

elrayle:
This is the final prep refactoring before adding ability to use ldpath with findterm.

* makes rdf_helper obsolete
* always converts id to URI before mapping results
* don’t filter graph if json-ld was requested


app/services/qa/linked_data/mapper/term_results_mapper_service.rb
```diff
@@ -0,0 +1,44 @@
...
+          #     {:uri=>[#<RDF::URI:0x3fcff54a829c URI:http://id.loc.gov/authorities/genreForms/gf2011026559>],
+          #      :id=>[#<RDF::Literal:0x3fcff4a367b4("gf2011026559")>],
+          #      :label=>[#<RDF::Literal:0x3fcff54a9a98("Science Fiction"@en)>],
+          #      :altlabel=>[#<RDF::Literal:0x3fcff54a9a98("Sci-fi"@en)>],
+          #      :broader=>[#<RDF::URI:0x3fcff54a8234 URI:http://id.loc.gov/authorities/genreForms/gf2014026339>],
+          #      :narrower=>[#<RDF::URI:0x3fcff54a8283 URI:http://id.loc.gov/authorities/genreForms/gf2014026551>],
+          #      :sameas=>[#<RDF::URI:0x3fcff54a8248 URI:http://id.loc.gov/authorities/names/n2010043281>]}
+          #   ]
+          def map_values(graph:, subject_uri:, predicate_map: nil)
+            graph_predicate_mapper_service.map_values(graph: graph, predicate_map: predicate_map, subject_uri: subject_uri)
```
elrayle: Uses existing GraphPredicateMapperService to map values for findterm results given a predicate map.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -43,34 +42,36 @@ def initialize(term_config)
       def find(id, language: nil, replacements: {}, subauth: nil, jsonld: false)
         raise Qa::InvalidLinkedDataAuthority, "Unable to initialize linked data term sub-authority #{subauth}" unless subauth.nil? || term_subauthority?(subauth)
         @language = language_service.preferred_language(user_language: language, authority_language: term_config.term_language)
-        url = authority_service.build_url(action_config: term_config, action: :term, action_request: normalize_id(id), substitutions: replacements, subauthority: subauth, language: @language)
+        @id = id
```
elrayle: This combined with the attr_reader makes id available anywhere in this class.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -43,34 +42,36 @@ def initialize(term_config)
...
+        url = authority_service.build_url(action_config: term_config, action: :term, action_request: normalize_id, substitutions: replacements, subauthority: subauth, language: @language)
```
elrayle: Unchanged except that no longer need to pass id to `normalize_id` since the id is available through an attr_reader.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -43,34 +42,36 @@ def initialize(term_config)
...
         Rails.logger.info "QA Linked Data term url: #{url}"
         load_graph(url: url)
         return "{}" unless full_graph.size.positive?
         return full_graph.dump(:jsonld, standard_prefixes: true) if jsonld
-        parse_term_authority_response(id)
+        filter_graph
```
elrayle: Filtering the graph was happening at the same time the full graph was read in.  If the full graph is being returned, there is no need to perform the filtering, so moved it to a separate method that is only called after it is determined we aren't going to pass back the full graph.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -43,34 +42,36 @@ def initialize(term_config)
...
+        results = map_results
+        convert_results_to_json(results)
```
elrayle: Based on single responsibility principle and to make consistent with the calls used in searchquery.rb, the mapping and json conversion happen in separate methods.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -79,82 +80,57 @@ def expects_uri?
           term_config.term_id_expects_uri?
         end
 
-        def extract_uri_from_id(id)
-          return @uri = id if expects_uri?
+        def uri
+          return @uri if @uri.present?
```
elrayle: Since what we really want is the uri, renamed this method to be consistent with ruby standard practice.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -79,82 +80,57 @@ def expects_uri?
...
+          return @uri = RDF::URI.new(id) if expects_uri?
```
elrayle: Fix bug -- Exception was thrown when `expects_uri? # true` because `@uri` was getting set to a `String` when it needs to be an `RDF::URI`.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -79,82 +80,57 @@ def expects_uri?
...
           @uri = graph_service.subjects_for_object_value(graph: @filtered_graph, predicate: RDF::URI.new(term_config.term_results_id_predicate), object_value: id.gsub('%20', ' ')).first
         end
 
         def preds_for_term
-          { required: required_term_preds, optional: optional_term_preds }
```
elrayle: Predicate map wants to be a flat hash, so collapsed the required and optional into a single hash.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -79,82 +80,57 @@ def expects_uri?
...
-          preds[:sameas] = term_config.term_results_sameas_predicate unless term_config.term_results_sameas_predicate.nil?
-          preds
+        def optional_preds
+          opt_preds = {}
+          opt_preds[:altlabel] = term_config.term_results_altlabel_predicate
+          opt_preds[:id] = term_config.term_results_id_predicate
+          opt_preds[:narrower] = term_config.term_results_narrower_predicate
+          opt_preds[:broader] = term_config.term_results_broader_predicate
+          opt_preds[:sameas] = term_config.term_results_sameas_predicate
+          opt_preds.delete_if { |_k, v| v.blank? }
```
elrayle: Instead of testing on every line setting optional predicates, simply remove any that are blank.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -79,82 +80,57 @@ def expects_uri?
...
+          return nil unless results.key?(key) || results[key].blank?
+          results[key]
         end
 
-        def select_json_result_for_id(json_results, id)
-          json_results.select! { |r| r[:uri].include? id } if json_results.size > 1
-          json_results.select! { |r| r[:uri].ends_with? id } if json_results.size > 1
-          json_results.first
+        def sort_literals(results, key)
+          return nil unless results.key? key
```
elrayle: This just got moved down the file and the parameter was changed from `consolidated_hash` (which no longer exists) to `results`.

lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -38,7 +38,7 @@ def search(query, language: nil, replacements: {}, subauth: nil, context: false)
         url = authority_service.build_url(action_config: search_config, action: :search, action_request: query, substitutions: replacements, subauthority: subauth, language: @language)
         Rails.logger.info "QA Linked Data search url: #{url}"
         load_graph(url: url)
-        results = process_results
+        results = map_results
```
elrayle: Just changed the method name to be more descriptive of what the method does.

PR #233, opened 2019-04-30
Add ldpath for findterm

elrayle:
Final PR refactoring findterm to add ldpath support for selected out results.

### DIFF
--- a/app/services/qa/linked_data/mapper/term_results_mapper_service.rb
+++ b/app/services/qa/linked_data/mapper/term_results_mapper_service.rb
@@ -4,6 +4,7 @@ module Qa
     module Mapper
       class TermResultsMapperService
         class_attribute :graph_ldpath_mapper_service, :graph_predicate_mapper_service
+        self.graph_ldpath_mapper_service = Qa::LinkedData::Mapper::GraphLdpathMapperService
         self.graph_predicate_mapper_service = Qa::LinkedData::Mapper::GraphPredicateMapperService
 
         class << self
@@ -11,6 +12,23 @@ module Qa
           # If a sort key is present, a subject will only be included in the results if it has a statement with the sort predicate.
           # @param graph [RDF::Graph] the graph from which to extract result values
           # @param subject_uri [RDF::URI] the uri of the term represented by the graph
+          # @param prefixes [Hash<Symbol><String>] URL map of prefixes to use with ldpaths
+          # @example prefixes
+          #   {
+          #     locid: 'http://id.loc.gov/vocabulary/identifiers/',
+          #     skos: 'http://www.w3.org/2004/02/skos/core#',
+          #   }
+          # @param ldpath [Hash<Symbol><String||Symbol>] value either maps to a ldpath in the graph or is :subject_uri indicating to use the subject uri as the value
+          # @example ldpath map
+          #   {
+          #     uri: :subject_uri,
+          #     id: 'locid:lccn :: xsd::string',
+          #     label: 'skos:prefLabel :: xsd::string',
+          #     altlabel: 'skos:altLabel :: xsd::string',
+          #     broader:  'skos:broader :: xsd::anyURI',
+          #     narrower: 'skos:narrower :: xsd::anyURI',
+          #     sameas: 'skos:exactMatch :: xsd::anyURI'
+          #   }
           # @param predicate_map [Hash<Symbol><String||Symbol>] value either maps to a predicate in the graph or is :subject_uri indicating to use the subject uri as the value
           # @example predicate map
           #   {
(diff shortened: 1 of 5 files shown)


app/services/qa/linked_data/mapper/term_results_mapper_service.rb
```diff
@@ -34,9 +52,27 @@ class << self
           #      :narrower=>[#<RDF::URI:0x3fcff54a8283 URI:http://id.loc.gov/authorities/genreForms/gf2014026551>],
           #      :sameas=>[#<RDF::URI:0x3fcff54a8248 URI:http://id.loc.gov/authorities/names/n2010043281>]}
           #   ]
-          def map_values(graph:, subject_uri:, predicate_map: nil)
-            graph_predicate_mapper_service.map_values(graph: graph, predicate_map: predicate_map, subject_uri: subject_uri)
+          def map_values(graph:, subject_uri:, prefixes: nil, ldpath_map: nil, predicate_map: nil)
+            if ldpath_map.present?
+              map_values_with_ldpath_map(graph: graph, ldpath_map: ldpath_map, prefixes: prefixes, subject_uri: subject_uri)
```
elrayle: Add in support for processing ldpath_map OR predicate_map.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -66,9 +66,21 @@ def filter_graph
         end
 
         def map_results
-          results_mapper_service.map_values(graph: @filtered_graph,
-                                            subject_uri: uri,
-                                            predicate_map: preds_for_term)
+          predicate_map = preds_for_term
+          ldpath_map = ldpaths_for_term
```
elrayle: Supports both ldpath_map (preferred) and predicate_map (deprecated).  Cannot mix ldpath and predicates.  A configuration must define results using one or the other.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -86,9 +98,26 @@ def uri
...
+          opt_ldpaths[:id] = term_config.term_results_id_ldpath
+          opt_ldpaths[:narrower] = term_config.term_results_narrower_ldpath
+          opt_ldpaths[:broader] = term_config.term_results_broader_ldpath
+          opt_ldpaths[:sameas] = term_config.term_results_sameas_ldpath
+          opt_ldpaths.delete_if { |_k, v| v.blank? }
+        end
+
         def preds_for_term
           label_pred_uri = term_config.term_results_label_predicate
-          raise Qa::InvalidConfiguration, "required label_predicate is missing in configuration for LOD authority #{authority_name}" if label_pred_uri.nil?
```
elrayle: Since the requirement now is that there be **either** predicate map OR ldpath map, the check and raising of an exception is moved up to the caller of this method.

spec/services/linked_data/mapper/term_results_mapper_service_spec.rb
```diff
@@ -2,56 +2,93 @@
...
+      sameas: ['http://cat.aii.caas.cn/concept/c_26308',
+               'http://lod.nal.usda.gov/nalt/20627',
+               'http://d-nb.info/gnd/4147072-2']
     }
   end
-  let(:subjects) { subject.map { |result| result[:uri].first.to_s } }
+
+  let(:results) do
+    results = {}
+    subject.each { |k, v| results[k] = v.map(&:to_s) }
```
elrayle: Rewrite RDF::Literals as strings because two instances of RDF::Literal with the same string value are not pass the `==` equality test, but their `.to_s` string values do pass equality tests.

spec/services/linked_data/mapper/term_results_mapper_service_spec.rb
```diff
@@ -2,56 +2,93 @@
 
 RSpec.describe Qa::LinkedData::Mapper::TermResultsMapperService do
   let(:graph) { Qa::LinkedData::GraphService.load_graph(url: 'http://local.data') }
-  let(:predicate_map) do
+  let(:subjects) { subject.map { |result| result[:uri].first.to_s } }
+
+  let(:expected) do
```
elrayle: Both the predicate_map test and the ldpath_map test use the same expected results.  Both approaches should give the same results.

lib/qa/authorities/linked_data/config/term_config.rb
```diff
@@ -72,7 +72,10 @@ def term_results_id_ldpath
       # Return results id_predicate
       # @return [String] the configured predicate to use to extract the id from the results
       def term_results_id_predicate
-        Config.predicate_uri(term_results, :id_predicate)
+        return @pred_id unless @pred_id.blank?
+        @pred_id = Config.predicate_uri(term_results, :id_predicate)
```
elrayle: A config can specify that either a simple id (e.g. 'sh12345') or the full uri will be used to find a term.  If the config indicates a simple id will be passed in, then it needs to be able to find the subject URI that has the ID as the object value.  To do this, it does a query...  `s? <id_pred> '_ID_'`.  To execute this query it needs to know the `<id_pred>`.  

When using a predicate map, the predicate map tells you the `<id_pred>`.  When using an ldpath map, the id predicate needs to be created from the prefixes and the ldpath map.

When `term_results_id_predicate` is called, it either returns the predicate from the predicate_map or calls `pred_id_from_ldpath` to generate the id predicate.

PR #239, opened 2019-05-07
do not discard blank nodes

elrayle:
The mapper ignores blank nodes when processing potential subject uris.  So it is not necessary to remove them in search_query.  Also, by leaving them in the graph, it allows extended context to be drawn from blank nodes.

### DIFF
--- a/lib/qa/authorities/linked_data/search_query.rb
+++ b/lib/qa/authorities/linked_data/search_query.rb
@@ -15,8 +15,8 @@ module Qa::Authorities
         @search_config = search_config
       end
 
-      attr_reader :search_config, :graph, :language, :access_time_s, :normalize_time_s
-      private :graph, :language, :access_time_s, :normalize_time_s
+      attr_reader :search_config, :full_graph, :filtered_graph, :language, :access_time_s, :normalize_time_s
+      private :full_graph, :filtered_graph, :language, :access_time_s, :normalize_time_s
 
       delegate :subauthority?, :supports_sort?, :prefixes, :authority_name, to: :search_config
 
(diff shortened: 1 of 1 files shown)


lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -58,7 +58,7 @@ def load_graph(url:)
         def normalize_results
           normalize_start_dt = Time.now.utc
 
-          @graph = graph_service.filter(graph: @graph, language: language, remove_blanknode_subjects: true)
+          @filtered_graph = graph_service.filter(graph: @full_graph, language: language)
           results = map_results
```
elrayle: This filtering is unnecessary.  [SearchResultsMapperService](https://github.com/samvera/questioning_authority/blob/master/app/services/qa/linked_data/mapper/search_results_mapper_service.rb#L68) ignores blank nodes when looking for potential subject uri results.

lib/qa/authorities/linked_data/search_query.rb
```diff
@@ -48,7 +48,7 @@ def search(query, language: nil, replacements: {}, subauth: nil, context: false,
         def load_graph(url:)
           access_start_dt = Time.now.utc
 
-          @graph = graph_service.load_graph(url: url)
+          @full_graph = graph_service.load_graph(url: url)
```
elrayle: I separated out full_graph from filtered_graph to be consistent with how these are treated in find_term.

lib/qa/authorities/linked_data/search_query.rb
elrayle: There are no test changes because removing here and ignoring in SearchResultsMapperService was redundant.

PR #240, opened 2019-05-08
reset uri with every call to find

elrayle:
Fixes a bug where the uri, once set, was used for all subsequent calls involving find by id.

### DIFF
--- a/lib/qa/authorities/linked_data/find_term.rb
+++ b/lib/qa/authorities/linked_data/find_term.rb
@@ -15,8 +15,8 @@ module Qa::Authorities
         @term_config = term_config
       end
 
-      attr_reader :term_config, :full_graph, :filtered_graph, :language, :id, :access_time_s, :normalize_time_s
-      private :full_graph, :filtered_graph, :language, :id, :access_time_s, :normalize_time_s
+      attr_reader :term_config, :full_graph, :filtered_graph, :language, :id, :uri, :access_time_s, :normalize_time_s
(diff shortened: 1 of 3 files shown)


spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -156,6 +161,15 @@
             .to eq ['headings beginning with the word [Scientific;] and subdivision [Science] under ethnic groups and individual wars, e.g. [World War, 1939-1945--Science]']
           expect(results['predicates']['http://www.w3.org/2004/02/skos/core#inScheme']).to eq ['http://id.loc.gov/authorities/subjects']
         end
+
+        it 'has correct primary predicate values for second request' do
+          expect(results[:uri]).to eq 'http://id.loc.gov/authorities/subjects/sh85118553'
+          expect(second_results[:uri]).to eq 'http://id.loc.gov/authorities/subjects/sh1234'
```
elrayle: This test fails before the fix.  The failure is only seen on the second find by id which is why this test checks `results` before checking `second_results`.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -80,6 +80,7 @@ def perform_normalization
           return full_graph.dump(:jsonld, standard_prefixes: true) if jsonld?
 
           filter_graph
+          extract_uri
```
elrayle: Extracting the uri here forces the uri to be updated for every call to find based on the current graph represented in `filter_graph`.

PR #249, opened 2019-06-04
Raise an exception if data can’t be normalized

### LINKED ISSUE #247: performance data for linked data term fetch fails if there are predicates is nil
### Description

If `results['predicates'].blank?`, then results['predicates'].size` raises an exception.

### Expected

If `results['predicates'].blank?`, #append_performance_data in find_term.rb should set...

```
performance = { predicate_count: 0, ...}
```

### Actual

If `results['predicates'].blank?`, #append_performance_data in find_term.rb raises...

```
exception Undefined method :size for NilClass
```

elrayle:
Closes #247

In this case, the data can’t be normalized because the requested ID does not match the ID in the graph for the term.  This prevents the extraction of the URI from the graph based on the ID.

With this fix, an exception is raised when it happens so the error can be reported back to users.  This fails more gracefully than the previous failure which just returned empty data for uri, id, and label.

### DIFF
--- a/app/controllers/qa/linked_data_terms_controller.rb
+++ b/app/controllers/qa/linked_data_terms_controller.rb
@@ -80,6 +80,10 @@ class Qa::LinkedDataTermsController < ::ApplicationController
                 "was not identified as a valid RDF format.  You may need to include the linkeddata gem."
     logger.warn msg
     render json: { errors: msg }, status: :internal_server_error
+  rescue Qa::DataNormalizationError => e
+    msg = "Data Normalization Error - #{e.message}"
+    logger.warn msg
+    render json: { errors: msg }, status: :internal_server_error
   end
 
   # Return all the information for a given term given a URI
--- a/lib/qa/authorities/linked_data/find_term.rb
+++ b/lib/qa/authorities/linked_data/find_term.rb
@@ -133,16 +133,26 @@ module Qa::Authorities
 
         def extract_uri
           return @uri = RDF::URI.new(id) if expects_uri?
-          @uri = graph_service.subjects_for_object_value(graph: @filtered_graph, predicate: RDF::URI.new(term_config.term_results_id_predicate), object_value: URI.unescape(id)).first
-          return @uri unless loc? && @uri.blank?
-          # for backward compatibility, if an loc id as passed in fails to extract the URI, try to adding a blank to the id
-          @uri = graph_service.subjects_for_object_value(graph: @filtered_graph, predicate: RDF::URI.new(term_config.term_results_id_predicate), object_value: loc_id).first
-          if @uri.present?
-            Qa.deprecation_warning(
-              in_msg: 'Qa::Authorities::LinkedData::FindTerm',
-              msg: 'Special processing of LOC ids is deprecated; id should be an exact match of the id in the graph'
-            )
-          end
+          extract_uri_by_id
+        end
+
+        def extract_uri_by_id
+          @uri = get_uri_from_graph_by_id(id)
+        rescue Qa::DataNormalizationError
+          raise unless loc?
+          @uri = get_uri_from_graph_by_id(loc_id)
+          Qa.deprecation_warning(
+            in_msg: 'Qa::Authorities::LinkedData::FindTerm',
+            msg: 'Special processing of LOC ids is deprecated; id should be an exact match of the id in the graph'
+          )
+          @uri
+        end
+
+        def get_uri_from_graph_by_id(gid)
+          @uri = graph_service.subjects_for_object_value(graph: @filtered_graph,
+                                                         predicate: RDF::URI.new(term_config.term_results_id_predicate),
+                                                         object_value: URI.unescape(gid)).first
+          raise Qa::DataNormalizationError, "Unable to extract URI based on ID: #{id}" if @uri.blank?
           @uri
         end
 
@@ -240,7 +250,8 @@ module Qa::Authorities
         end
 
         def append_performance_data(results)
-          performance = { predicate_count: results['predicates'].size,
+          pred_count = results['predicates'].present? ? results['predicates'].size : 0
+          performance = { predicate_count: pred_count,
                           fetch_time_s: access_time_s,
                           normalization_time_s: normalize_time_s,
                           total_time_s: (access_time_s + normalize_time_s) }
--- a/spec/lib/authorities/linked_data/find_term_spec.rb
+++ b/spec/lib/authorities/linked_data/find_term_spec.rb
@@ -90,90 +90,131 @@ RSpec.describe Qa::Authorities::LinkedData::FindTerm do
             .to include('Cornell University', 'Ithaca (N.Y.). Cornell University', "Kornel\\xCA\\xB9skii universitet",
                         "K\\xCA\\xBBang-nai-erh ta hs\\xC3\\xBCeh")
         end
+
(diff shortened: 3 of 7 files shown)


app/controllers/qa/linked_data_terms_controller.rb
```diff
@@ -80,6 +80,10 @@ def show # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
                 "was not identified as a valid RDF format.  You may need to include the linkeddata gem."
     logger.warn msg
     render json: { errors: msg }, status: :internal_server_error
+  rescue Qa::DataNormalizationError => e
+    msg = "Data Normalization Error - #{e.message}"
+    logger.warn msg
+    render json: { errors: msg }, status: :internal_server_error
```
elrayle: Added exception to be used for reporting problems normalizing the RDF to JSON.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -133,16 +133,26 @@ def expects_uri?
...
-          return @uri unless loc? && @uri.blank?
-          # for backward compatibility, if an loc id as passed in fails to extract the URI, try to adding a blank to the id
-          @uri = graph_service.subjects_for_object_value(graph: @filtered_graph, predicate: RDF::URI.new(term_config.term_results_id_predicate), object_value: loc_id).first
-          if @uri.present?
-            Qa.deprecation_warning(
-              in_msg: 'Qa::Authorities::LinkedData::FindTerm',
-              msg: 'Special processing of LOC ids is deprecated; id should be an exact match of the id in the graph'
-            )
-          end
+          extract_uri_by_id
```
elrayle: This is mostly a reorganizing to make the code less cluttered and DRY.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -133,16 +133,26 @@ def expects_uri?
...
+            msg: 'Special processing of LOC ids is deprecated; id should be an exact match of the id in the graph'
+          )
+          @uri
+        end
+
+        def get_uri_from_graph_by_id(gid)
+          @uri = graph_service.subjects_for_object_value(graph: @filtered_graph,
+                                                         predicate: RDF::URI.new(term_config.term_results_id_predicate),
+                                                         object_value: URI.unescape(gid)).first
+          raise Qa::DataNormalizationError, "Unable to extract URI based on ID: #{id}" if @uri.blank?
```
elrayle: This is the new bit that raised the error if the ID can't be found in the graph as an object of a triple.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -133,16 +133,26 @@ def expects_uri?
...
-            Qa.deprecation_warning(
-              in_msg: 'Qa::Authorities::LinkedData::FindTerm',
-              msg: 'Special processing of LOC ids is deprecated; id should be an exact match of the id in the graph'
-            )
-          end
+          extract_uri_by_id
+        end
+
+        def extract_uri_by_id
+          @uri = get_uri_from_graph_by_id(id)
```
elrayle: This is the normal way for getting the URI from the ID.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -133,16 +133,26 @@ def expects_uri?
...
+        rescue Qa::DataNormalizationError
+          raise unless loc?
+          @uri = get_uri_from_graph_by_id(loc_id)
```
elrayle: Trying a second time with the LOC specially processed ID.  Only happens when the authority is accessing LOC directly.

lib/qa/authorities/linked_data/find_term.rb
```diff
@@ -240,7 +250,8 @@ def predicates_with_subject_uri(expected_uri) # rubocop:disable Metrics/MethodLe
         end
 
         def append_performance_data(results)
-          performance = { predicate_count: results['predicates'].size,
+          pred_count = results['predicates'].present? ? results['predicates'].size : 0
+          performance = { predicate_count: pred_count,
```
elrayle: This original error was that `results['predicates']` was nil and causing `.size` to raise an exception.  This prevents that.  This really shouldn't be needed, but if it were to happen again in another circumstance, it would fail more gracefully.

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -90,90 +90,131 @@
...
 
     context 'in LOC authority' do
       context 'term found' do
-        before do
-          stub_request(:get, 'http://id.loc.gov/authorities/subjects/sh85118553')
-            .to_return(status: 200, body: webmock_fixture('lod_loc_term_found.rdf.xml'), headers: { 'Content-Type' => 'application/rdf+xml' })
-          stub_request(:get, 'http://id.loc.gov/authorities/subjects/sh1234')
-            .to_return(status: 200, body: webmock_fixture('lod_loc_second_term_found.rdf.xml'), headers: { 'Content-Type' => 'application/rdf+xml' })
-        end
+        context 'when id requires special processing for <blank> in id' do
```
elrayle: This is mostly a refactor moving the stubs closer to the tests that use them for readability purposes only.

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -90,90 +90,131 @@
...
+          let(:results_without_blank) { lod_loc.find('sh85118553', subauth: 'subjects') }
+
+          it 'extracts correct uri' do
+            expect(results_without_blank[:uri]).to eq 'http://id.loc.gov/authorities/subjects/sh85118553'
+          end
         end
 
-        it 'extracts correct uri when loc id does not have blank' do
-          expect(results_without_blank[:uri]).to eq 'http://id.loc.gov/authorities/subjects/sh85118553'
+        context "ID in graph doesn't match ID in request URI" do
```
elrayle: This is the new test that checks for the DataNormalizationError exception.

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -90,90 +90,131 @@
             .to include('Cornell University', 'Ithaca (N.Y.). Cornell University', "Kornel\\xCA\\xB9skii universitet",
                         "K\\xCA\\xBBang-nai-erh ta hs\\xC3\\xBCeh")
         end
+
+        context "ID in graph doesn't match ID in request URI" do
+          before do
```
elrayle: Tests for DataNormalizationError with the OCLC_FAST authority.  Catches the case where special processing of the ID is NOT done.

spec/lib/authorities/linked_data/find_term_spec.rb
```diff
@@ -90,90 +90,131 @@
...
+            stub_request(:get, 'http://id.loc.gov/authorities/subjects/sh85118553')
+              .to_return(status: 200, body: webmock_fixture('lod_loc_term_found.rdf.xml'), headers: { 'Content-Type' => 'application/rdf+xml' })
+          end
+
+          let(:results_without_blank) { lod_loc.find('sh85118553', subauth: 'subjects') }
+
+          it 'extracts correct uri' do
+            expect(results_without_blank[:uri]).to eq 'http://id.loc.gov/authorities/subjects/sh85118553'
+          end
```
elrayle: ENDS refactor that moved stubs closer to tests that use them

PR #261, opened 2019-07-31
Remove CircleCI ruby/rails version defaults

botimer:
Re: samvera/maintenance#20
Re: samvera/maintenance#21


jrgriffiniii:
These failures appear to be related to the WorldCat server errors which appeared for earlier test jobs on builds such as https://circleci.com/workflow-run/bba233e6-582f-4127-a825-92746790501d

PR #264, opened 2019-08-08
allow error from RDF::Graph to be parsed with or without paren around code

elrayle:
### Problem

Multiple webmock tests, but not all, were randomly failing on CCI only.  These were first seen when PR #259 was merged into master, but were not caused by the PR.  They appeared to be related to dependency changes or infrastructure change at CCI.  

After extensive investigation (many thanks to @botimer and @jrgriffiniii), it was determined that the addition of rest-client gem caused an error to surface.  The error occurs because when rest-client gem is present, RDF::Graph.load method returns its error messages in a different form.  Sometimes the error message includes parentheses around the error code (i.e., "... (503)") and sometimes not (i.e., "... 503").  QA parses the error code from the error message to pass back a more meaningful error to the caller.  

### Solution

QA was doing a positional parse expecting the message to be consistent.  It was updated to use regex which is less fragile.

### A Better Solution

Ultimately, the fix would be for RDF::Graph.load to...
1) minimally return the error message in the same format every time
2) improve exceptions by throwing specific exceptions based on the error code that can be captured in the calling application

### DIFF
--- a/app/services/qa/linked_data/graph_service.rb
+++ b/app/services/qa/linked_data/graph_service.rb
@@ -106,12 +106,10 @@ module Qa
             end
           end
 
+          # process ioerror_code from RDF::Graph.load whether the code is in parentheses (i.e. "... (404)"), or not (i.e. "... 404")
           def ioerror_code(e)
             msg = e.message
-            return 'format' if msg.start_with? "Unknown RDF format"
-            a = msg.size - 4
-            z = msg.size - 2
-            msg[a..z]
+            msg[/(\(?)(\d\d\d)(\)?)$/, 2]
           end
       end
     end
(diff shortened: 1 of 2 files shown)


elrayle:
I tried...
* pinning webmock to 3.6.0 (latest is 3.6.2)
* pinning net-http-persistent to 3.0.1 (latest is 3.1.0)
* forcing the test app cache to regenerate

Rather or not the tests fail for any particular combination of Rails and Ruby versions appears to be random.  Best I can tell, which tests fail are the same each time failures occur.

NOTE: There have been not recent changes to tests.  And the tests that are failing have been passing consistently for a long time.  Some of the failing tests haven't been touched in 2 years.

PR #284, opened 2019-11-11
add language processing to ldpath service

elrayle:
The ldpath_service was not honoring requests for languages.  With this change, if languages are passed into the ldpath_program method and the property being requested has a language marker in it, the program will create a request for each language and for all untagged literals.

The ldpath_execute method will recombine these values into language tagged RDF::Literals.

Without this, the language processing in QA is unable to properly sort multiple values within a single language.


jrgriffiniii:
Apologies for any annoyance, but this is to please document and confirm that https://github.com/samvera-labs/ldpath/pull/18 is actually a blocking PR for merging these proposed changes.

PR #299, opened 2019-12-03
update Discogs authority to support pagination parameters

tworrall:


lib/qa/authorities/discogs/discogs_utils.rb
```diff
@@ -55,6 +55,28 @@ def bf_role_type_object
         "http://id.loc.gov/ontologies/bibframe/Role"
       end
 
+      def format(tc)
+        return 'json' unless tc.params.key?('format')
```
elrayle: Nice putting these in utils.

PR #321, opened 2020-08-13
Changes LoC authorities to use https

### LINKED ISSUE #320: Library of Congress now redirecting HTTP requests to HTTPS
LOC recently starting requiring `https:` for all `id.loc.gov` requests. (Identifiers are still `http:`, but if you access the site using `http:`, you’ll get a `302 Moved Permanently` redirect to the `https:` version). If QA isn’t already respecting redirects (and probably even if it is), it should probably be updated to use `https:` on all LOC calls.

rotated8:
Fixes #320

LoC is now redirecting `http` requests to `https`, and QA is not following redirects, so applications would see HTTP 500 errors, because the response would not parse as JSON. This commit fixes that issue by changing all the LoC urls to use `https`.


elrayle:
QA doesn't parse out the ID.  For fetch, the ID is passed in...

```
http://localhost:3000/qa/show/loc/subjects/sh2008121753
```

where sh2008121753 is the id passed to QA for fetching data.

PR #323, opened 2020-08-14
Fixes issue limiting LoC results to an authority

### LINKED ISSUE #320: Library of Congress now redirecting HTTP requests to HTTPS

rotated8:
Fixes #320 (again).

While the URL to search LoC authorities has changed to https, the URL for authorities (used to limit results to that authority) has not.

### DIFF
--- a/lib/qa/authorities/loc_subauthority.rb
+++ b/lib/qa/authorities/loc_subauthority.rb
@@ -70,18 +70,18 @@ module Qa::Authorities::LocSubauthority
   private
 
     def vocab_base_url
-      "cs%3Ahttps%3A%2F%2Fid.loc.gov%2Fvocabulary%2F"
+      "cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fvocabulary%2F"
     end
 
     def authority_base_url
-      "cs%3Ahttps%3A%2F%2Fid.loc.gov%2Fauthorities%2F"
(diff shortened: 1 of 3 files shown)


rotated8:
@elrayle `cs` appears to be a non-documented feature. In the [docs](https://id.loc.gov/techcenter/searching.html), `scheme` seems to do the same thing. This bears out for the search results. Compare [using cs](https://id.loc.gov/search/?q=Emory&q=cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fnames&format=json) and [using scheme](https://id.loc.gov/search/?q=Emory+scheme%3Ahttp%3A%2F%2Fid.loc.gov%2Fauthorities%2Fnames&format=json) in a search for Emory against the LCNAF, and [using cs](https://id.loc.gov/search/?q=france+cs%3Ahttp%3A%2F%2Fid.loc.gov%2Fvocabulary%2Fcountries&format=json) and [using scheme](https://id.loc.gov/search/?q=france+scheme%3Ahttp%3A%2F%2Fid.loc.gov%2Fvocabulary%2Fcountries&format=json) in a search for France against the MARC List for Countries.

Should we change how we limit results?

PR #342, opened 2021-11-04
add optional URI to locally defined yml file authorities

elrayle:
Also, refactor to move tests for locally defined authorities under the directory where they live in code.


lib/qa/authorities/local/file_based_authority.rb
```diff
@@ -8,13 +8,17 @@ def initialize(subauthority)
     def search(q)
       r = q.blank? ? [] : terms.select { |term| /\b#{q.downcase}/.match(term[:term].downcase) }
       r.map do |res|
-        { id: res[:id], label: res[:term] }.with_indifferent_access
+        qa_res = { id: res[:id], label: res[:term] }.with_indifferent_access
+        qa_res[:uri] = res[:uri] if res.key?(:uri)
+        qa_res
```
jcoyne, line 13: This works, but there is a more shorthand way to write this:
```ruby
{ id: res[:id], label: res[:term], uri: res[:uri] }.compact.with_indifferent_access
```

jcoyne (approved):

lib/qa/authorities/local/file_based_authority.rb
elrayle: Good to know.  I'll update.

jcoyne (approved):

PR #348, opened 2022-01-05
add pagination service and use it for yml file authorities

elrayle:
This implementation follows json-api recommendations with:
* response[‘data’] = the results
* response[‘meta’][‘page’] = pagination statistics (i.e., page_offset, page_limit, actual_page_size, total_num_found)
* response[‘links’] = set of links for pagination (i.e., self, first, prev, next, last)
* response[‘errors’] = any errors with requested page_offset or page_limit (only included if there are errors)

Reference:
* [Pagination](https://jsonapi.org/format/#fetching-pagination) section of [Latest JsonAPI spec](https://jsonapi.org/format/)
* [Pagination Links](https://jsonapi.org/examples/#pagination) example

NOTE: For backward compatibility, if the format is JSON (default) and neither page_offset or page_limit are set, then only the results, including all results, will be returned.

### DIFF
--- a/app/services/qa/pagination_service.rb
+++ b/app/services/qa/pagination_service.rb
@@ -1,6 +1,83 @@
 module Qa
-  # Provide pagination processing that authority modules can use to respond to
-  # requests for paginated results.
+  # Provide pagination processing used to respond to requests for paginated results.
+  #
+  # <b>Defaults for page_offset and page_limit:</b> (see example responses under #build_response)
+  #
+  # format == :json
+  #
+  # * if neither page_offset nor page_limit is passed in, then... (backward compatible)
+  #   * returns results as an Array<Hash>
+  #   * returns all results
+  #   * page_offset is "1"
+  #   * page_limit is total_num_found
+  #
+  # * if either page_offset or page_limit is passed in, then...
+  #   * returns results as an Array<Hash>
+  #   * returns a page of results
+  #   * default for page_offset is "1"
+  #   * default for page_limit is DEFAULT_PAGE_LIMIT (i.e., 10)
+  #
+  # format == :jsonapi
+  #
+  # * response is always in the jsonapi format
+  # * results are always paginated
+  # * default for page_offset is "1"
+  # * default for page_limit is DEFAULT_PAGE_LIMIT (i.e., 10)
+  #
+  # <b>How page_offset is calculated for pagination links:</b>
+  #
+  # expected page boundaries
+  #
+  # * Expected page boundaries are always calculated starting from page_offset=1
+  #   and the current page_limit.  The page boundaries will include the page_offsets
+  #   that cover all results.  For example, page_limit=10 with 36 results will have
+  #   page boundaries 1, 11, 21, 31.
+  #
+  # self url
+  #
+  # * The self url always has the page_offset for the current page, which defaults
+  #   to 1 if not passed in.
+  #
+  # first page url
+  #
+  # * The first page url always has page_offset=1.
+  #
+  # last page url
+  #
+  # * The last page url always has page_offset equal to the last of the expected page
+  #   boundaries regardless of the passed in page_offset.  For the example where
+  #   page_limit=10 with 36 results, the last page will always have page_offset=31.
+  #
+  # prev page url
+  #
+  # * Previous' page_offset is calculated from the passed in page_offset whether or
+  #   not it is on an expected page boundary.
+  #
+  # * For prev, page_offset = passed in page_offset - page_limit || nil if calculated as < 1
+  #   * when current page_offset (e.g. 1) is less than page_limit (e.g. 10), then page_offset
+  #     for prev will be nil (e.g. 1 - 10 = -9 which is < 1)
+  #   * when current page_offset is an expected page boundary (e.g. 21), then
+  #     page_offset for prev will also be a page boundary (e.g. 21 - 10 = 11
+  #     which is an expected page boundary)
+  #   * when current page_offset is not on an expected page boundary (e.g. 13), then
+  #     page_offset for prev will not be on an expected page boundary (e.g. 13 - 10 = 3
+  #     which is not an expected page boundary)
+  #
+  # next page url
+  #
+  # * Next's page_offset is calculated from the passed in page_offset whether or
+  #   not it is on an expected page boundary.
+  #
+  # * For next, page_offset = passed in page_offset + page_limit || nil if calculated > total number of results found
+  #   * when current page_offset (e.g. 31) is greater than total number of results (e.g. 36),
+  #     then page_offset for next will be nil (e.g. 31 + 10 = 41 which is > 36)
+  #   * when current page_offset is an expected page boundary (e.g. 21), then
+  #     page_offset for next will also be a page boundary (e.g. 21 + 10 = 31
+  #     which is an expected page boundary)
+  #   * when current page_offset is not on an expected page boundary (e.g. 13), then
+  #     page_offset for next will not be on an expected page boundary (e.g. 13 + 10 = 23
+  #     which is not an expected page boundary)
+  #
   class PaginationService # rubocop:disable Metrics/ClassLength
     # Default page_limit to use if not passed in with the request.
     DEFAULT_PAGE_LIMIT = 10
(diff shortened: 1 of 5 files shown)


spec/services/pagination_service_spec.rb
```diff
@@ -0,0 +1,689 @@
...
+      context 'with page_offset missing' do
+        context 'and page_limit is missing' do
+          it 'returns the first DEFAULT_PAGE_LIMIT (10) records' do
+            expect(response["data"]).to match_array(results[0..9])
+          end
+
+          it "sets meta['page'] stats" do
+            expect(response["meta"]["page"]["page_offset"]).to eq "1"
+            expect(response["meta"]["page"]["page_limit"]).to eq described_class::DEFAULT_PAGE_LIMIT.to_s
+            expect(response["meta"]["page"]["actual_page_size"]).to eq described_class::DEFAULT_PAGE_LIMIT.to_s
```
cjcolvar, line 378: How about using a hard coded 10 here?  It seems like you've done that elsewhere and I think it would be easier to read.
```suggestion
            expect(response["meta"]["page"]["page_limit"]).to eq "10"
            expect(response["meta"]["page"]["actual_page_size"]).to eq "10"
```

spec/services/pagination_service_spec.rb
```diff
@@ -0,0 +1,689 @@
...
+            expect(response["meta"]["page"]["total_num_found"]).to eq "36"
+          end
+
+          it "sets prev link to nil and next link to next page" do
+            next_url = "#{base_url}#{url_path}?q=term&format=jsonapi&page_limit=10&page_offset=11"
+            expect(response['links']['prev']).to be_nil
+            expect(response['links']['next']).to eq next_url
```
cjcolvar, line 385: I think you can use `second_page` here:
```suggestion
            expect(response['links']['prev']).to be_nil
            expect(response['links']['next']).to eq second_page
```

spec/services/pagination_service_spec.rb
```diff
@@ -0,0 +1,689 @@
...
+      context 'and results do not include a full page' do
+        context 'and there are several results in range' do
+          let(:query_string) { 'q=term&format=jsonapi&page_offset=31&page_limit=10' }
+
+          before do
+            params[:page_offset] = "31"
+            query_params[:page_offset] = "31"
+          end
+
+          it 'returns json api response with a partial page of results starting at 11th result' do
```
cjcolvar, line 505:
```suggestion
          it 'returns json api response with a partial page of results starting at 31th result' do
```

spec/services/pagination_service_spec.rb
```diff
@@ -0,0 +1,689 @@
...
+
+          it "sets meta['page'] stats" do
+            expect(response["meta"]["page"]["page_offset"]).to eq "31"
+            expect(response["meta"]["page"]["page_limit"]).to eq "10"
+            expect(response["meta"]["page"]["actual_page_size"]).to eq "6"
+            expect(response["meta"]["page"]["total_num_found"]).to eq "36"
+          end
+
+          it "sets self to last page" do
+            expect(response['links']['self']).to eq last_page
```
cjcolvar, line 517: Why is this `last_page`?  I don't understand why this case is handled differently than the next one where the page_offset is 36.

spec/services/pagination_service_spec.rb
```diff
@@ -0,0 +1,689 @@
...
+          it 'does not include errors' do
+            expect(response.key?('errors')).to eq false
+          end
+        end
+
+        context 'and page_offset is start of fourth page (last page)' do
+          let(:page_offset) { "31" }
+
+          it 'returns json api response third page of results' do
+            expect(response["data"]).to match_array(results[30..39])
```
cjcolvar, line 663: Interesting that this works if results only goes to 35.  Maybe it would still be worth updating to be clear?
```suggestion
            expect(response["data"]).to match_array(results[30..35])
```

cjcolvar (commented):
This is looking really good!  I have a few small cleanup suggestions and a couple questions that might lead to some reworking.

spec/services/pagination_service_spec.rb
elrayle: This is why this worked...
```
>> results[30..39]
[{:id=>"31", :label=>"term 31"}, {:id=>"32", :label=>"term 32"}, {:id=>"33", :label=>"term 33"}, {:id=>"34", :label=>"term 34"}, {:id=>"35", :label=>"term 35"}, {:id=>"36", :label=>"term 36"}]
```

But I definitely agree that it is better to update it to the expected range.

spec/services/pagination_service_spec.rb
elrayle: The `page_offset` (i.e., 31) in this example falls on an expected page boundary (e.g. 1, 11, 21, 31).  And 31 is the start of the `last_page` allowing the `last_page` url to be used for `response['links']['self']`.

In the next example, `page_offset` is 36 which is not on an expected page boundary.  So the defined `last_page` is not the page that was requested.  This requires `self_url` to be defined in the spec.

You didn't ask, but one of the decision points on behavior that I wrangled with is the definition of `last_page` when the `page_offset` is not on an expected page boundary.  I decided that `prev` and `next` would be calculated off the passed in `page_offset` (i.e. `page_offset -/+ page_limit`), but `last_page` would always be the last page calculated from `page_offset=1` which always puts it on an expected page boundary.

cjcolvar: :+1: That makes total sense and looking at this again I can't understand what I was thinking before.

app/services/qa/pagination_service.rb
```diff
@@ -0,0 +1,513 @@
...
+      # @return [Integer] The first record to include in the response data as
+      #     requested as long as it is an Integer; otherwise, returns nil.
+      def validated_request_page_offset
+        @page_offset_error = false
+        offset = Integer(requested_page_offset)
+        return offset if offset == 1
+        @page_offset_error = ERROR_OUT_OF_RANGE_TOO_SMALL if offset < 1
+        @page_offset_error = ERROR_OUT_OF_RANGE_TOO_LARGE if offset > total_num_found
+        offset
+      rescue
```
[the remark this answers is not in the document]
elrayle: DONE

spec/controllers/terms_controller_spec.rb
```diff
@@ -122,6 +122,78 @@ def search(_arg1, _arg2)
...
+          it "sets meta['page'] stats" do
+            get :search, params: params
+            json_api_response = JSON.parse(response.body)
+            expect(json_api_response["meta"]["page"]["page_offset"]).to eq "6"
+            expect(json_api_response["meta"]["page"]["page_limit"]).to eq "5"
+            expect(json_api_response["meta"]["page"]["actual_page_size"]).to eq "5"
+            expect(json_api_response["meta"]["page"]["total_num_found"]).to eq total_num_found
+          end
+
+          it 'sets links with next set to nil' do
```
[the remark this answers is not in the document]
elrayle: DONE

spec/services/pagination_service_spec.rb
elrayle: I added documentation for the class describing this process.  Let me know if it is clear enough.

app/services/qa/pagination_service.rb
```diff
@@ -0,0 +1,513 @@
...
+        @page_limit_error = ERROR_NOT_INTEGER
+        nil
+      end
+
+      # @return [Integer] The default to use when page_limit isn't specified.
+      # @note To maintain backward compatibility, the limit will be all results
+      #   if format is json and neither page_limit nor page_offset were specified.
+      def default_page_limit
+        return total_num_found unless json_api? || page_offset_specified? || page_limit_specified?
```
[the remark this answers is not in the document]
elrayle: I started off having `page_limit=total_num_found` if `format==:json` and `page_limit` is not passed in regardless of whether `page_offset` was passed in.  As I thought about it more, it seemed to me that if `page_offset` is passed in, then it implies a desire for pagination, in which case, the default `page_limit` should be used.

app/services/qa/pagination_service.rb
```diff
@@ -0,0 +1,563 @@
...
+      end
+
+      # @return [Integer] The max number of records to include in response data as
+      #     requested as long as it is a positive Integer; otherwise, returns nil.
+      def validated_request_page_limit
+        @page_limit_error = false
+        limit = Integer(requested_page_limit)
+        @page_limit_error = ERROR_OUT_OF_RANGE_TOO_SMALL if limit < 1
+        limit.positive? ? limit : nil
+      rescue
```
cjcolvar, line 460: Match the change above to only rescue specific errors.
```suggestion
      rescue ArgumentError
```

app/services/qa/pagination_service.rb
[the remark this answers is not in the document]
cjcolvar: I think that totally makes sense and provides the same pagination behavior across both the `:json` and `:json_api` formats.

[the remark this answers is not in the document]
elrayle: Here is an example of how it would look with 36 results if the default for `:json` was to default `page_limit=total_num_found`…

- page_offset=1 - returns all [1..36]
- page_offset=11 - returns [11..36]
- page_offset=21 - returns [21..36]
- page_offset=31 - returns [31-36]

As opposed to using the default `page_limit` (i.e., 10)…

- page_offset=1 - returns [1..10]
- page_offset=11 - returns [11..20]
- page_offset=21 - returns [21..30]
- page_offset=31 - returns [31-36]

cjcolvar (approved):
This looks great! :tada:

PR #350, opened 2022-02-07
Revert "wrap initialization in reloader.to_prepare to avoid autoload deprecation warning in some cases"

cjcolvar:
Reverts samvera/questioning_authority#349

Based upon the issue [I was seeing in Hyrax](https://github.com/samvera/questioning_authority/pull/349#issuecomment-1030326627) trying to use the new release of QA, I think it would be good to revert this change, cut a new release, and open an issue to take a deeper look at how to resolve this.


elrayle (approved):
Thanks for this.  Issue coming soon to describe next steps.

PR #402, opened 2025-05-30
Remove unnecessary setting

eddierubeiz:
It turns out QA does follow redirects within the same host; we won't be needing this setting.


cjcolvar (approved):
Looks good.

PR #410, opened 2025-12-17
Replace custom deprecation gem with ActiveSupport::Deprecation

mark-dce:
**ISSUE**
The [deprecation gem](https://rubygems.org/gems/deprecation) is no longer being maintained (see https://github.com/projectblacklight/blacklight/pull/2793).

**RESOLUTION**
Update to use native Rails deprecation functionality.

### DIFF
--- a/lib/qa.rb
+++ b/lib/qa.rb
@@ -25,10 +25,22 @@ module Qa
     @config
   end
 
+  def self.deprecator
+    @deprecator ||= deprecator_instance
+  end
+
+  # TODO: refactor when support for Rails < 7.1.0 is dropped
+  def self.deprecator_instance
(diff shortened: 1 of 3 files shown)


jrochkind:
looks good, thanks!

jrochkind:
@mark-dce notes:

1.  Deprecation warnings had previously been silenced in all test runs. Now they honor the configuration of the parent application or test environment variables, So, other apps using QA may see more deprecation warnings in test suites.

2. QA still runs tests against unsupported Ruby and Rails versions, so I had to add some janky conditional logic for when to use ActiveSupport::Deprecations and when to just log directly. If we were willing to drop support for the older Ruby & Rails versions, things could be a little cleaner.

PR #418, opened 2026-06-05
Pin `multi_json` when json is version `2.19.x` 

eddierubeiz:
This fixes the tests, which were failing because  a `multi_json 1.19.x` / `json 2.19.x` interaction causes
 `RDF::Graph#dump(:jsonld, standard_prefixes: true)` to raise:
 `NoMethodError: undefined method 'except' for JSON::Ext::Generator::State` in:
* lib/qa/authorities/linked_data/find_term.rb
* lib/qa/authorities/discogs/discogs_translation.rb

### DIFF
--- a/spec/test_app_templates/Gemfile.extra
+++ b/spec/test_app_templates/Gemfile.extra
@@ -29,7 +29,10 @@ group :development do
   # 1) this passes on all versions:
   #gem "multi_json", "1.17.0"
 
-  # 2) let's try this now:
-  gem "multi_json", "!= 1.19.1"
+  # 2) this gives us 1.19.0, which also fails.
+  #gem "multi_json", "!= 1.19.1"
+
+  # 3) now we try this:
+  gem "multi_json", ">= 1.17", "< 1.19"
 
 end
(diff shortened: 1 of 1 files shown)


jrochkind:
Thank you!!

Fixing it by not including multi-json in ruby 4...   do we understand why multi_json is being included at all (in GEmfile.extra only?), and what the effects are of not including it?  I guess multi_json is not updated for ruby 4?

Since it's just Gemfile.extra.... it just goes into _generated_ app for testing...  I'm not sure it even effects anyone except generated app? 

I wonder if we should just remove multi_json altogether, instead of the conditional leaving mysterious complexity for future devs?

eddierubeiz:
Oh, we're not excluding the gem `multi-json`; we're just preventing CI from using version `1.19.x` for Ruby 3.0 and 3.1. Could also prevent the app itself for using `1.19.x, too`. Not sure if I'd put that in the same PR; suggestions welcome.

jrochkind:
Oh I see that now, misread. Approved!

jrochkind (approved):

PR #419, opened 2026-06-08
move benchmark require inside task so only required when needed

### LINKED ISSUE #414: `benchmark` needs to be listed in gemspec
> /Users/jrochkind/.gem/ruby/3.4.8/gems/qa-5.16.0/lib/tasks/mesh.rake:1: warning: /Users/jrochkind/.rubies/ruby-3.4.8/lib/ruby/3.4.0/benchmark.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 4.0.0.
You can add benchmark to your Gemfile or gemspec to silence this warning.
Also please contact the author of qa-5.16.0 to request adding benchmark into its gemspec.

Or not used in qa? Haven't looked to see what we are using it for!

jrochkind:
Instead of on application boot (loading all rakefiles). As benchmark is removed from default gems in ruby 4, requiring it triggers a warning in late ruby 3, or will require you to add as a dependency in ruby 4. For now, we just move require inside so only an issue if you actually run task instead of on booting any app.

Closes #414

### DIFF
--- a/lib/tasks/mesh.rake
+++ b/lib/tasks/mesh.rake
@@ -1,9 +1,9 @@
-require 'benchmark'
-
 namespace :qa do
   namespace :mesh do
     desc "Import MeSH terms from the file $MESH_FILE, it will update any terms which are already in the database"
     task import: :environment do
+      require 'benchmark' # may require you to add benchmark dependency in ruby 4+
+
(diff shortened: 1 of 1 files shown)


cjcolvar (approved):
👍

PR #421, opened 2026-06-16
Replace deprecated `Bundler.with_clean_env`

eddierubeiz:
`Bundler.with_clean_env` was deprecated in Bundler 2.1 and completely removed in Bundler 4.0 .

Let's replace it with `Bundler.with_unbundled_env` .


cjcolvar (approved):
👍 Thanks for doing this cleanup!

PR #423, opened 2026-06-17
Update version file to 5.17.0

eddierubeiz:
I am preparing [a new release](https://github.com/samvera/questioning_authority/releases/tag/5.17.0) using Github's new release tool, but we still need to update the version file.

### DIFF
--- a/lib/qa/version.rb
+++ b/lib/qa/version.rb
@@ -1,3 +1,3 @@
 module Qa
-  VERSION = "5.16.0".freeze
+  VERSION = "5.17.0".freeze
 end


jrochkind:
I don't know anything about that tool, but if it means you already published something with wrong version, that woudl be bad!

jrochkind (approved):

jrochkind:
OK, eddie I think you created a release on github, but you have not actually released to rubygems. 

Also, as you noticed, the release on github has a version tag pointed at a version of the code with the wrong version. 

Please fix the version tag in git to be pointed at the right thing *before* you release to rubygems. Release to rubygems is also required.

